PR #7, opened 2015-08-19
Add julia 0.4 to REQUIRE

joshday:
The README says UnicodePlots is developed for 0.3 and 0.4.  I'd like to try this out on 0.4, but right now Pkg.add("UnicodePlots") doesn't work.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.3
+julia 0.4


IainNZ:
@joshday thats not how it works. `julia 0.3` means it works for Julia 0.3 or greater. Have you done Pkg.update?

IainNZ:
It was only registered a few days ago.

joshday:
Ah, my mistake.  I was reading the documentation too quickly...Pkg.update did the trick.  Thanks.

PR #8, opened 2015-08-19
Add missing 'io' argument.

MichaelHatherly:
When printing to something other than ``STDOUT`` the missing ``io`` argument was messing up the output.

(Awesome package btw!)


Evizero:
good catch! Thanks

PR #29, opened 2015-09-24
Fix deprecation warnings on Julia 0.4

jiahao:
Now requires Compat.jl to maintain compatibility with Julia 0.3


Evizero:
Sweet, dealing with AbstractString and FloatingPoint was actually on my TODO list.
Thanks!

PR #30, opened 2015-09-26
select the number of histogram bins automatically

bicycle1885:
This will automatically select the number of bins based on Sturges' rule.
This rule is used in the `Base.hist` function.


Evizero:
Awesome, thanks. I'll test it right after lunch

Evizero:
Thanks!

bicycle1885:
you're welcome :smiley:

PR #43, opened 2015-12-06
fix demo image background

waldyrious:
The background of this image had some artifacts, I cleaned it up a bit. The others may have the same issue, I didn't check. This one was more evident because of the white "dots" under the top left plot.


Evizero:
thanks

Issue #51, opened 2016-07-31
InexactError() when plotting


lopezm94:
When trying to plot `Array{Float64, N}`, sometimes I get the following error:

```julia
ERROR: InexactError()
 in Plot at /home/USER/.julia/v0.4/UnicodePlots/src/plot.jl:151
```

*There are no NaN or Inf values*

Evizero:
can you give me a little more details? What was the command your executed, and what were your trying to do? What is `N` in your case?

lopezm94:
It happens with N in [1,2] (but mostly 2), the calls are almost identical, here is the matrix case:

```julia
julia> plot = scatterplot([left],[miny],xlim=[left,iters],ylim=[miny,maxy],title=title,name=name)
ERROR: InexactError()
 in Plot at /home/juan/.julia/v0.4/UnicodePlots/src/plot.jl:150

#Variale (name, type, value)
left Int64 1
iters Int64 34
miny Float64 -0.8794276565174709
maxy Float64 1.2796649117521434e218
title ASCIIString 
```

I see this tends to happen when the difference between miny and maxy is big

Evizero:
I can reproduce this. I shall investigate

Evizero:
The "issue" is with `round`. It is caused because I try to convert floats that are integers (like 1.0) into integers so that the axis labels are a little less noisy. In this case though the number is too big to be stored as an Int64 which causes the `InexactError`.

Evizero:
I think it should be fixed by replacing the four

```julia
min_y_str = string(isinteger(min_y) ? round(Int, min_y, RoundNearestTiesUp) : min_y)
```

lines at https://github.com/Evizero/UnicodePlots.jl/blob/master/src/plot.jl#L148-L151 with a bounds check

```julia
min_y_str = string((isinteger(min_y) &&  typemin(Int) < min_y < typemax(Int)) ? round(Int, min_y, RoundNearestTiesUp) : min_y)
```

I can't fix it right now but I shall do it in the next few days. If you want you could also submit a PR with these changes

PR #52, opened 2016-07-31
Fix InexactError() bug

lopezm94:
Fix for #51, tested it with random inputs for a while a worked like a charm.


lopezm94:
Is the only way to not get the error is to install the master branch? or can the PR somehow propagate to a previous release?

Evizero:
can you maybe add your previous call that caused the error into the (very messy and unprofessional) test file?

As soon as I merge I will tag a new version, so no you won't have to checkout the master

lopezm94:
I couldn't help but chuckle at your test cases, I tried to mimic your test theme as much as I could. 

I also added some more round checks in common.jl just in case.

lopezm94:
I don't really understand coveralls, but if you want I can try to fix that

Evizero:
Thanks! No don't mind coverall. One day I will have to think of actual sensible testing (Turns out testing a terminal plotting library is not that easy)

Evizero:
https://github.com/JuliaLang/METADATA.jl/pull/5764

PR #56, opened 2016-09-13
fixes for 0.5

tbreloff:
As suggested in https://github.com/Evizero/UnicodePlots.jl/issues/55 this is a fix for the dep warnings in 0.5.  Also there were some "method overwritten" warnings for the `annotate!` method, and I think I fixed those while keeping the same behavior (please check!)

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.4
+Compat 0.9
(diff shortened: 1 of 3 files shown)


Evizero:
Thanks, but I don't want to introduce compat just to remove it again with the next change. I'd rather go full 0.5 with the next tag. I think we are at the last RC. will take care of it once 0.5 is released

tbreloff:
Ok no problem.

PR #63, opened 2017-02-15
Fixes for Julia 0.6

andreasnoack:


Evizero:
Thanks for the effort!

andreasnoack:
Thanks for the package.

src/graphics/bargraphics.jl
```diff
@@ -1,3 +1,5 @@
+if VERSION < v"0.6-"
```
tkelman: please be more specific with VERSION cutoffs like this, `where` isn't going to work for all 0.6-dev versions

Evizero: what would be a way to find a reasonable lower bound? any pointer where to start looking. Is it related to the merge of https://github.com/JuliaLang/julia/pull/18457 ?

tkelman: yes, check out the merge commit there and run contrib/commit-name.sh

PR #68, opened 2017-08-09
Fixes calls to `histogram(...)` with arrays of `Float32`

JohnHBrock:
`float_round_log10` now returns same float type it was given.

This (at the very least) fixes calls to `histogram(...)` with arrays of `Float32`: previously, calling `histogram` with an array of `Float32` would crash with:
```jl
ERROR: LoadError: TypeError: float_round_log10: in typeassert, expected Float64, got Float32
```


Evizero:
thanks!

JohnHBrock:
@Evizero Any chance we could get a new release now that this is merged?

Evizero:
right, thanks for reminding. will do

Evizero:
https://github.com/JuliaLang/METADATA.jl/pull/10667

PR #69, opened 2017-08-12
updated stemplots

alexhallam:
I wanted to give this another shot. A few new features have been added. Let me know what needs to happen and I will get on it.

### DIFF
--- a/src/interface/.zshrc
+++ b/src/interface/.zshrc
@@ -1,0 +1,1 @@
+export LC_ALL=en_US.UTF-8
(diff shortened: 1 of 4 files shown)


Evizero:
A cool! Will take a look asap (monday morning probably)

src/interface/.zshrc
```diff
@@ -0,0 +1 @@
+export LC_ALL=en_US.UTF-8
```
Evizero: we shouldn't add this file to the repo

Evizero:
looks good to me, nice. we probably should have one or two calls in the test file just to make sure the code executes

PR #74, opened 2017-12-22
Replace all occurrences of Vector by AbstractVector in plot commands

### LINKED ISSUE #73: Support for other AbstractArray
Thanks for the package, it is awesome. I tried to create a barplot today with a `MVector` containing the values and realized that the function are only defined in terms of `Vector`. Would it be a good idea to generalize the signature of the plot commands to work with any `AbstractArray` or `AbstractVector`?

juliohm:
Fix #73 

The outputs in the test suite looked correct to me, but I appreciate if you can double check.

### DIFF
--- a/src/interface/densityplot.jl
+++ b/src/interface/densityplot.jl
@@ -88,23 +88,23 @@ See also
 `Plot`, `scatterplot`, `DensityCanvas`
 """
 function densityplot{F<:Real,R<:Real}(
-        x::Vector{F},
-        y::Vector{R};
+        x::AbstractVector{F},
+        y::AbstractVector{R};
         color::Symbol = :white,
         args...)
-    X = convert(Vector{Float64}, x)
-    Y = convert(Vector{Float64}, y)
+    X = convert(AbstractVector{Float64}, x)
+    Y = convert(AbstractVector{Float64}, y)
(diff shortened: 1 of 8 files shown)


src/interface/densityplot.jl
```diff
@@ -88,23 +88,23 @@ See also
...
-        x::Vector{F},
-        y::Vector{R};
+        x::AbstractVector{F},
+        y::AbstractVector{R};
         color::Symbol = :white,
         args...)
-    X = convert(Vector{Float64}, x)
-    Y = convert(Vector{Float64}, y)
+    X = convert(AbstractVector{Float64}, x)
+    Y = convert(AbstractVector{Float64}, y)
```
Evizero: this doesn't look like a good idea. let's leave this as `Vector` for `convert`

src/interface/densityplot.jl
```diff
@@ -88,23 +88,23 @@ See also
...
```
Evizero: same here (leave `Vector`)

juliohm:
Oh yeah, that wasn't intended, will fix it tomorrow morning. Calling the
day here...

juliohm:
Done :+1:

Evizero:
Thanks!

PR #79, opened 2018-06-29
Julia 07 fixes

Paulms:
Fix Julia v0.7 warnings and deprecations


Evizero:
Thanks for tackling this! 

That said, I'm afraid it will take me a bit before i can review and merge this, as I have not started the migration process and thus don't quite know much about 0.7 yet

Paulms:
Ok, no problem, take your time.

KristofferC:
I wouldn't recommend commiting the Manifest file here.

Paulms:
Ok, I remove the Manifest file, but now I get some  MethodError: no method matching haskey(::Nothing, ::String) related with _get_deps! on Travis. It seems extrange that for Julia nightly travis works (since I assume nigthly is also Julia 0.7 beta). All test are passed on my local machine so I am not sure if it is a temporary problem with travis.

KristofferC:
Beta is significantly behind nightly at this point. I would focus on getting test passing on nightly for now.

Evizero:
Thanks!

PR #82, opened 2018-09-15
implemented arbitrary scale transformations for barplots

ExpandingMan:
This implements arbitrary scale transformations for the ordinate of bar plots (including histograms).  I did this because, though I love this package, the inability to make log plots was severely limiting its usefulness for me.

Obviously there are all sorts of features regarding scale transformations which would be nice to have which I have still not implemented.  I really did not go too crazy here, but my hope is that what's here will nevertheless be a good starting point and provide a sorely needed basic functionality for some simple use cases.  Of course, it was already possible to implement some simple transformations by doing, e.g. `barplot(labels, log10.(x))`, but with this PR it is possible to do this while retaining the original (untransformed) labels.  For example:
```julia
julia> histogram(randn(10^5), transform=log10)
               ┌────────────────────────────────────────┐ 
   (-4.5,-4.0] │ 1                                      │ 
   (-4.0,-3.5] │▇▇▇▇▇▇▇▇ 12                             │ 
   (-3.5,-3.0] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 101                     │ 
   (-3.0,-2.5] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 488               │ 
   (-2.5,-2.0] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 1624          │ 
   (-2.0,-1.5] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 4375       │ 
   (-1.5,-1.0] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 9298    │ 
   (-1.0,-0.5] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 14910  │ 
    (-0.5,0.0] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 19116 │ 
     (0.0,0.5] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 19292 │ 
     (0.5,1.0] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 14963  │ 
     (1.0,1.5] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 9106     │ 
     (1.5,2.0] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 4433       │ 
     (2.0,2.5] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 1672          │ 
     (2.5,3.0] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 457                │ 
     (3.0,3.5] │▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 127                    │ 
     (3.5,4.0] │▇▇▇▇▇▇▇▇▇▇ 22                           │ 
     (4.0,4.5] │▇▇▇▇ 3                                  │ 
               └────────────────────────────────────────┘ 
```
This PR is of course a little elaborate for providing that ability alone, but again, my hope is that it can be used as a starting point to add more features at least for barplots.

If you approve and are interested in merging this, I suggest we leave the first PR pretty simple and if we do more, save it for future updates.  Thanks!

### DIFF
--- a/src/UnicodePlots.jl
+++ b/src/UnicodePlots.jl
@@ -13,7 +13,10 @@ export
             BlockCanvas,
             AsciiCanvas,
             DotCanvas,
-        BarplotGraphics,
+        BarplotGraphicsArea,
+            BarplotGraphics,
+            BarplotTransform,
+            barplotgraphics,
 
     pixel_width,
     pixel_height,
--- a/src/graphics/bargraphics.jl
+++ b/src/graphics/bargraphics.jl
@@ -1,4 +1,6 @@
-mutable struct BarplotGraphics{R<:Number} <: GraphicsArea
+abstract type BarplotGraphicsArea <: GraphicsArea end
+
+mutable struct BarplotGraphics{R<:Number} <: BarplotGraphicsArea
     bars::Vector{R}
     color::Symbol
     width::Int
@@ -18,8 +20,18 @@ mutable struct BarplotGraphics{R<:Number} <: GraphicsArea
     end
 end
 
+mutable struct BarplotTransform{R′<:Number,R<:Number,T<:Function} <: BarplotGraphicsArea
+    graphics::BarplotGraphics{R}
+    bars′::Vector{R′}
+    max_freq′::R′
+    transform::T
+end
+
 nrows(c::BarplotGraphics) = length(c.bars)
 ncols(c::BarplotGraphics) = c.width
+
+nrows(c::BarplotTransform) = length(c.graphics.bars)
+ncols(c::BarplotTransform) = c.graphics.width
 
 function BarplotGraphics(
         bars::AbstractVector{R},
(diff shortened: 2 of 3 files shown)


Evizero:
Sorry it is taking me so long to get to this. Its mainly because I don't have any quick feedback. I am not an immediate fan of solving this with a new type for some reason. That said I can't think of anything constructive to say without first spending some time really thinking about this.

It will take me another week at least to get to this. sorry for the delay

Evizero:
Ok so I thought about it a bit. I think its worth compromising here to get this functionality out to those who really want it (e.g. you). That said I am not fully ready to commit to long time support of the exact specifics. So my suggestion would be to merge this PR as is with the exception of leaving the `export` statements as they were before. This way I leave the option option to internally breaking changes.

ExpandingMan:
I agree with you that at present this seems like extreme overkill for supporting this functionality.  This wasn't much work, so I'm in no way attached to it and am willing to help with an alternative if there's a good one.

Let me suggest one: one could, of course simply pass a transformed vector into a bar plot.  One could add an option to `histogram` that causes this to happen.  The only thing that would be missing from this is that all the labels be `f(x)` rather than `x`.  So, another option is to add code which comprehensively supports alternative labeling.  The way things are right now, it's not obvious how to do that.  Most types would probably need to get at least one new field.

Just to give you an idea of what I'm ultimately after here: for me personally, I could probably use this package exclusively for 95% of my exploratory plotting if all axes supported log, and that would be really cool, because I very much like to do everything in the terminal :slightly_smiling_face:

PR #83, opened 2018-10-09
Update readme.md density plot example

anderson15:
Added dots before plus for new broadcasting.  This is an amazing package!!! Thanks!


Evizero:
thanks!

PR #88, opened 2018-12-17
add heatmaps

rjkat:
Heatmaps have their own canvas type, with a single block of colour per pixel. Instead of blending adjacent colours using `|` (which leads to strange colours), the latest colour written to a pixel is the winner.

Heatmaps are made from a matrix `z`, with the `xscale` and `yscale` parameters determining the scale for the axis labels. There is currently no colourbar to show the z scale. Probably lots of improvements which can be made - feel free to comment/criticise and I will do my best to address.


Evizero:
I am really sorry that I forgot about this.

I like the idea but I think that if we really go for a heatmap we should do the approach used by https://github.com/JuliaImages/ImageInTerminal.jl . Instead of a single block per character it uses two stacked blocks per character and sets the two colors individually by using foreground and background colors in the terminal. This way the aspect ratio is also proper.

However, I don't think the lookupcanvas would work for this, so it would be a more complicated implementation

rjkat:
@Evizero I have implemented the foreground/background stuff, which results in much nicer heatmaps :) It's added a dependency on Crayons.jl (which could probably be removed if necessary) and probably could use some further tweaks. For example, plots with odd numbers of rows aren't handled nicely. Would be interested to hear your thoughts.

rjkat:
Hey. Sorry for all the commits. I can squash them into one if you'd prefer. I was busy making the default sizing nice and adding a colorbar with styling options. Also fixing corner cases like zero-sized arrays. Here is a screenshot of the heatmap being used to do a spectrogram of some audio, which is what I'm mainly using it for.

spectrogram screenshot

Evizero:
wow this looks amazing. Thanks a lot for doing this.

I see that I can get around to review this PR somewhere this week.

rjkat:
Not sure how best to integrate this, but it's now possible to call heatmap directly on an array of RGB values and have it shown in the terminal. Similar to [https://github.com/JuliaImages/ImageInTerminal.jl](ImageInTerminal.jl) but nowhere near as fully-featured. I didn't want to add a dependency on ColorTypes to UnicodePlots, so instead of using proper types it's guessing based on the presence of `r`, `g` and `b` fields. Can either delete it, or leave it in but not documented. 
Screen Shot 2019-07-15 at 4 28 30 pm

rjkat:
Any more thoughts on this? I'd love to get it into the package, so more than happy to accommodate requests for changes to be made.

Evizero:
I am sorry i dropped the ball on this one. I went ahead and merged it into master. I'll wait a bit before I tag to let people try it out.

Thanks again for all that effort

PR #119, opened 2020-05-01
fix tests

dpo:
Tests fail for me when run locally. This PR fixes the barplot tests.

The spy tests also fail. Are the reference plots outdated?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,8 +17,9 @@ StatsBase = "0.32"
 julia = "1"
 
 [extras]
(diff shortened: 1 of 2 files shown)


dpo:
I supposed the syntax `@inferred <type> <call>` was introduced after Julia 1.1. The test fails for me on Julia 1.4. What do you suggest?

Evizero:
So does this PR fix it for Julia 1.4+ ? We could merge and bump the requirements in that case

t-bltg:
Superseded by https://github.com/Evizero/UnicodePlots.jl/pull/136.

PR #121, opened 2020-05-23
allow latest version of StatsBase.jl

bkamins:
It would be also great if a new version of the package could be registered if this gets merged. Thank you!


ericphanson:
Bump :)

andreasnoack:
@Evizero Would you be able to make a patch release with this?

Evizero:
should now work I think

PR #127, opened 2020-10-26
added support for vectors containing NaN in lineplot

bcsj:
Added a couple of lines which allows lineplot to work with vectors containing NaN values, e.g. 
image


Evizero:
thanks!

PR #128, opened 2020-11-11
Squeeze array before making histogram

rkurchin:
Thanks for a super useful package!

I have some data that, as read in by DelimitedFiles, gives me a 2000×1 Array{Float64,2}. Intuitively, I think the `histogram` function should be able to handle this, but it borks unless I do `histogram(data[:,1])`. Obviously this is an easy enough fix, but the stacktrace does not make it **super** obvious that that is the source of the issue, so I thought it might make sense to just squeeze the singleton dimensions by default, as I've done here.

I tried to add a test too, but when I ran the suite I got some type inference issues with another unrelated test in `tst_barplot.jl` (seems this is corroborated [here](https://github.com/Evizero/UnicodePlots.jl/pull/119)) so I wasn't able to verify that the test runs properly because when I tried to run just the histogram ones I fell into a morass of other package dependency issues :/


Evizero:
Hi! Thanks for the fix. Mhm the CI setup must be broken as I dont see any test output here. Would be nice to have some other opinions or even people trying this as am a bit out of touch with the code

t-bltg:
@rkurchin, can you update this PR ?

johnnychen94 (approved):

test/tst_histogram.jl
```diff
@@ -1,5 +1,6 @@
 seed!(RNG, 1337)
 x = randn(RNG, 10000)
+x2 = randn(RNG, (1, 10000, 1, 1))
```
johnnychen94, line 3:
```suggestion
x2 = reshape(x, (1, 10000, 1, 1))
```

t-bltg: Ha, yeah, even better, reshape in the test.

PR #133, opened 2021-04-13
Migrate from Travis to GitHub Actions

DilumAluthge:
1. Migrate from Travis to GitHub Actions
2. Update the TagBot configuration


DilumAluthge:
@Evizero Looks like there are a lot of test failures. Any idea what's going on?

Evizero:
is it working locally? looks to me like some fileio incompat

DilumAluthge:
The tests also fail for me locally.

Do you know what versions of FileIO, AstroIO, etc. the tests are compatible with?

Evizero:
I am afraid not. I would appreciate the help here

DilumAluthge:
The issue is with the ReferenceTests package. You have reference tests in `.txt` files, but when ReferenceTests tries to load them with FileIO, it interprets the file as being something other than text.

Probably this needs some fixes upstream in ReferenceTests and/or FileIO.

waldyrious:
Text should actually be the default assumption for ReferenceTests — see https://github.com/JuliaTesting/ReferenceTests.jl/issues/27 (/cc @christopher-dG, @oxinabox).

Also, https://github.com/JuliaTesting/ReferenceTests.jl/pull/19 and https://github.com/JuliaTesting/ReferenceTests.jl/issues/73 could potentially be related as well (/cc @juliohm, @johnnychen94).

johnnychen94:
Looks like the tests are broken a long time ago. I'm fully scheduled this April but I can take an investigation in May. @Evizero It would be quite convenient if you can send me an invitation.

PR #135, opened 2021-05-16
`heatmap`: add `zlim` kwarg for setting color range

IanButterworth:
Also fixes some docs for default values given `[0,0] == (0.,0.)` is `false`

image

### DIFF
--- a/src/interface/heatmap.jl
+++ b/src/interface/heatmap.jl
@@ -15,8 +15,8 @@ Usage
     heatmap(z::AbstractMatrix; title = "", width = 0, height = 0,
             xlabel = "", ylabel = "", zlabel = "", labels = true,
             border = :solid, colormap = :viridis, colorbar_border = :solid, show_colorbar = :true,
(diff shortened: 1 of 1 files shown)


johnnychen94:
This package hasn't switched to Github Action CI because of the legacy test code #133. I'm currently not available to investigate it so I'll just _assume_ that your code works. (The change looks good to me)

Given that CI is not set up, I prefer to not tag a new version. But if you need it, I can do this.

src/interface/heatmap.jl
```diff
@@ -118,6 +122,10 @@ function heatmap(z::AbstractMatrix; xlim = (0., 0.), ylim = (0., 0.), xoffset =
         noextrema = false
     catch
     end
+    if zrange != (0., 0.)
```
johnnychen94, line 125: Just for consistency, maybe `zlim` here?

IanButterworth: Yeah, that would be more consistent

PR #136, opened 2021-05-17
Bring back CI and upgrade to Julia 1.6

IanButterworth:
Progress:
- Switched over to github actions CI
- Created new test environment
- Updated reference tests (needs careful review!! I just blanket updated them without understanding whether it was correct to do so)
- Added a `force_raw_txt` kwarg to `ReferenceTests.@reference_test` in https://github.com/JuliaTesting/ReferenceTests.jl/pull/91, as a few of the reference test `.txt` files were being accidentally detected by FileIO to be `Gadget2` format 

Locally with the above PR, tests get to the barplot tests and fail on the `@inferred` because `Plot` is inferred, but `Plot{BarplotGraphics{Real}}` is returned

Fix https://github.com/Evizero/UnicodePlots.jl/issues/138
Fix https://github.com/Evizero/UnicodePlots.jl/issues/134
Close https://github.com/Evizero/UnicodePlots.jl/pull/119


t-bltg:
> Updated reference tests (needs careful review!! I just blanket updated them without understanding whether it was correct to do so)

I believe there is a change in the random sequences (we observed this in `Plots.jl`), so let's use a stable rng from `StableRNGs`.

@IanButterworth, the corresponding PR on your branch is here: https://github.com/IanButterworth/UnicodePlots.jl/pull/1.

IanButterworth:
Great. This means the reference images need updating again? Feel free to PR that update also

t-bltg:
> Great. This means the reference images need updating again? Feel free to PR that update also

Thanks, yes, I can generate the refs images again.

Can you maybe add me as a collaborator to your fork ? That would allow me to push directly into this branch, and not annoy you with more PRs.

IanButterworth:
> Can you maybe add me as a collaborator to your fork
done 👍🏻

t-bltg:
@IanButterworth , I think this PR is ready. Do you mind editing your first post by adding:
```
Fix https://github.com/Evizero/UnicodePlots.jl/issues/138.
Fix https://github.com/Evizero/UnicodePlots.jl/issues/134.
Close https://github.com/Evizero/UnicodePlots.jl/pull/119.
```

for Github to link fixed issues ?

@Evizero, can you review this PR, so that we get a running CI ?

Project.toml
```diff
@@ -15,10 +15,3 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 Crayons = "4.0"
 StatsBase = "0.32, 0.33"
 julia = "1"
```
johnnychen94, line 17: I see you have moved the test targets to `test/Project.toml` and that makes this package not compatible with old Julia versions. Given that it's about to make Julia 1.6 the next LTS release, I'm okay with this change.

```suggestion
julia = "1.2"
```

src/interface/histogram.jl
```diff
@@ -127,8 +127,8 @@ function histogram(
 end
 
 function histogram(x; bins = nothing, closed = :left, kw...)
-    if bins != nothing
-        Base.depwarn("The keyword parameter `bins` is deprecated, use `nbins` instead", :histogram)
+    if bins !== nothing
+        @warn "The keyword parameter `bins` is deprecated, use `nbins` instead"
```
johnnychen94, line 131: Hmmm, any reason for this change? I think the meaning of `depwarn` is quite clear here.

Users can switch to show the depwarn with `julia --depwarn=yes` before upgrading their codes.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
...
+          - ubuntu-latest
+        version:
+          # - '1.0'
+          # - '1.1'
+          - '1.2'
+          - '1.3'
+          - '1.4'
+          - '1.5'
+          - '1' # automatically expands to the latest stable 1.x release of Julia
+          - 'nightly'
```
johnnychen94, line 29: When the test passes, we can maybe switch to: (windows and macOS CI resources are quite expensive and I think it's enough to only have 1.x test on these).

```suggestion
      matrix:
        version: ['1.2', '1', 'nightly']
        os: [ubuntu-latest]
        arch: [x64]
        include:
          - os: windows-latest
            version: '1'
            arch: x64
          - os: macOS-latest
            version: '1'
            arch: x64
```

johnnychen94 (approved):
I didn't check the reference files. Hopefully, that's all because of the change of random generator. As long as the CI passes, I think it's good to merge.

src/interface/histogram.jl
t-bltg: Sorry was not familiar with `depwarn`. This chang was to fix the test here: https://github.com/Evizero/UnicodePlots.jl/blob/9630e2024a00c08b1f438c4907e036161b91a570/test/tst_histogram.jl#L82.

If tests are run with `--depwarn=yes` in the CI I'm gonna restore that, and add a message in the test file.

.github/workflows/ci.yml
t-bltg: Yes, I agree, it isn't necessary to cover all the version and burn tons of CI credits.

src/interface/histogram.jl
johnnychen94: By default `pkg> test` is triggered with `depwarn=yes` mode. https://github.com/JuliaLang/Pkg.jl/pull/1763

Project.toml
johnnychen94: For compatibility consideration, I personally prefer to still manage `[targets.test]` entries and not bump `julia` compat until we really need some new features that can't be easily patched. For instance, the `test/Project.toml` and something like `isnothing` can be perfectly written in another way and thus aren't good reasons to bump the `julia` compat.

Just my two cents.

src/interface/histogram.jl
t-bltg: Thanks, I've added a message in the test files to reduce confusions, since running with `include("test/runtests.jl")` (not `] test UnicodePlots`) does fails the test unless `--depwarn=yes` is specified.

.github/workflows/ci.yml
johnnychen94: The idea of my matrix setup is to:

- provide a test on lower-bound version `1.2` (as we declared in `Project.toml`). I believe this is necessary since it usually catches many unexpected compatibility bugs.
- only test windows and macOS platform on Julia 1.x version.
- `1.2` and `nightly` is only tested on Ubuntu.

This should be sufficient enough to catch most issues while use as little window/macOS CI credits as possible.

t-bltg: Thanks for the explanation, I will change that.

t-bltg:
@johnnychen94, do you have the ability to add me as a collaborator on this repo ? And avoid `workflow awaiting approval` on all CI related PRs ?

johnnychen94:
You can submit a trivial PR to master, e.g, add an empty line to README, and I can quickly merge it to work around this.

t-bltg:
> If @Evizero wants, I think we can move this package to JuliaPlots

I'm also in favor of this to ease maintenance, but it is of course @Evizero 's decision.

t-bltg:
The build failure on `1.2` for `test_canvas.jl` (`test/references/canvas/ascii_print.txt`) is odd.

**1.2**
```
┌ Debug: norm(x1)=7.795622224798846 norm(y1)=5.596882487442018 norm(x2)=17.157938577345448 norm(y2)=18.75298612996726
```

**1**
```
┌ Debug: norm(x1)=7.795622224798846 norm(y1)=5.596882487442018 norm(x2)=17.157938577345448 norm(y2)=18.75298612996726
```

So this isn't induces by the RNG change.

johnnychen94:
I can reproduce this locally and it's clearly not small differences:

image

image

There must be some places not working here.

t-bltg:
> I can reproduce this locally and it's clearly not small differences:
> 
> image
> 
> image
> 
> There must be some places not working here.
> 
> I would suggest starting from references generated by Julia 1.0, and then gradually upgrade to Julia 1.6.
> 
> FWIW, there's a `by` keyword that you can use to pass to `test_references` so that we can custom the equality check. For example, to skip leading/trailing spaces:
> 
> ```julia
> function test_equal(ref::String, actual::String)
>     all(zip(split(ref, r"\n|\r\n"), split(actual, r"\n|\r\n"))) do (x, y)
>         # line by line comparison without checking the leading/trailing spaces
>         strip(x) == strip(y)
>     end
> end
> function test_ref(reference, actual)
>     @test_reference(reference, actual, render = BeforeAfterFull(), format = "TXT", by=test_equal)
> end
> ```

Thanks, on which julia version ? I cannot reproduce this locally.

johnnychen94:
Julia 1.2 macOS. If you want a quick Julia version switch you might be interested in jill.py, I use it a lot to test packages on multiple Julia versions.

Project.toml
t-bltg: I'd prefer also to run the CI under the current LTS which is `1.0.5`.

t-bltg:
Interesting: on `1.2` and `1.6`, `v = c.grid[x,row] + 1` is correct but `lookup_decode[v]` has different behaviors across versions .

**MWE**

Copy lines 1-97 of https://github.com/Evizero/UnicodePlots.jl/blob/master/src/canvas/asciicanvas.jl in a julia repl:

**1.0.5 or 1.2.0**
```julia
julia> VERSION
v"1.0.5"
julia> ascii_decode[429]
'|': ASCII/Unicode U+007c (category Sm: Symbol, math)
```

**1.6.0 or up**
```julia
julia> VERSION
v"1.6.2"
julia> ascii_decode[429]
'^': ASCII/Unicode U+005E (category Sk: Symbol, modifier)
```

t-bltg:
I'm getting close, now `test/tst_spy.jl` fails only on `1.0`:

spy

t-bltg:
Finally, all tests pass :tada: !

@IanButterworth and @johnnychen94 I'll let you review this while I'm taking a :coffee: .

t-bltg:
I've checked visually all the references images using:

 ```bash
root="$HOME/Downloads/UnicodePlots.jl"

for f in $(find $root/test/references -name '*.txt'); do
  echo; echo
  echo "== local($f) =="
  cat $f
  echo
  url=https://raw.githubusercontent.com/Evizero/UnicodePlots.jl/master/${f#$root/}
  echo "== remote($url) =="
  wget -q -O - $url
done
```

, and everything looks good (no changes apart from random numbers, as expected).

File: src/common.jl [hunk 4]
```diff
@@ -71,6 +71,19 @@ function extend_limits(vec, limits)
     (limits == (0.,0.) || limits == [0.,0.]) ? plotting_range_narrow(mi, ma) : (mi, ma)
 end
 
+sort_by_keys(dict::Dict) = collect(sort(collect(dict), by=x->x[1]))
```
johnnychen94, line 74:
```suggestion
sort_by_keys(dict::Dict) = sort!(collect(dict), by=x->x[1])
```

src/common.jl
```diff
@@ -71,6 +71,19 @@ function extend_limits(vec, limits)
...
+
+function sorted_keys_values(dict::Dict; k2s=true)
```
johnnychen94, line 76: Just a question: is this sorting an expected operation for barplot and boxplot? Is it because dictionaries are not ordered?

johnnychen94 (approved):
@t-bltg This looks great! Thank you for the detective work and bringing back CI!

Just two small comments and with that resolved I think we can merge it.

src/common.jl
t-bltg: I added sorting, because iterating over test dicts in https://github.com/Evizero/UnicodePlots.jl/blob/326a4eb04279548891928cbb92868e0c51f0cf27/test/tst_barplot.jl#L20 has different behavior between `1.0` and `1.6` causing the comparison against reference files to fail in `1.0`.

I don't know if there is a better solution ?

johnnychen94: It seems doesn't change the overall behavior of `barplot`, i.e., :

```julia
barplot(["A", "B"], [1, 2])
barplot(["B", "A"], [2, 1])
```

generates different results.

t-bltg: That's expected, it doesn't not use `Dict`s here.

PR #140, opened 2021-07-20
Allow integer binning

### LINKED ISSUE #139: UnitRange has no field step
```julia
julia> using UnicodePlots, StatsBase

julia> a = fit(Histogram, rand(10), 0:3)
Histogram{Int64, 1, Tuple{UnitRange{Int64}}}
edges:
  0:3
weights: [10, 0, 0]
closed: left
isdensity: false

julia> histogram(a)
ERROR: type UnitRange has no field step
```

Moelf:
fix #139


Moelf:
image

this is more complicated than I though

PR #141, opened 2021-08-08
Allow irregularly spaced histogram bins

aminnj:
Removes the assumption that `StatsBase.Histogram` edges are a StepRange, which then allows:

```julia
julia> using UnicodePlots, StatsBase

julia> h = fit(Histogram, 10 .^ randn(10^4), [0.0,0.1,1,10,100]);

julia> UnicodePlots.histogram(h)
                  ┌                                        ┐
   [  0.0,   0.1) ┤▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 1589
   [  0.1,   1.0) ┤▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 3423
   [  1.0,  10.0) ┤▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 3441
   [ 10.0, 100.0) ┤▇▇▇▇▇▇▇▇▇▇▇▇▇ 1345
                  └                                        ┘
                                  Frequency

```


Evizero:
thanks! could you okace add a test for this new behaviour? Also does it break any existing tests? i guess CI is broken here

src/interface/histogram.jl
```diff
@@ -90,12 +90,13 @@ function histogram(
         kw...)
     edges, counts = hist.edges[1], hist.weights
     labels = Vector{String}(undef, length(counts))
-    binwidth = typeof(edges.step) == Float64 ? edges.step : edges.step.hi
+    binwidths = diff(edges)
```
johnnychen94, line 93: Since we don't currently have CI or test backed, so just a check, is `length(diff(edges)) == length(counts)`?

aminnj: This should always be true since the `Histogram` object itself requires this relationship:
```julia
julia> StatsBase.Histogram([1,2,3],[1,2,3])
ERROR: Histogram edge vectors must be 1 longer than corresponding weight dimensions
```

aminnj:
Even in `master`, I see most test cases fail because the bin counts are slightly different (seed related?). Is this expected?

When running the tests with this PR, there is one extra "failure":

```diff
index b9a4717..dd03729 100644
--- a/test/references/histogram/float32.txt
+++ b/test/references/histogram/float32.txt
@@ -1,6 +1,6 @@
-                  ┌                                        ┐
-   [0.0 , 0.05  ) ┤▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 1
-   [0.05, 0.1   ) ┤ 0
-   [0.1 , 0.15  ) ┤▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 2
-                  └                                        ┘
-                                  Frequency
\ No newline at end of file
+                ┌                                        ┐
+   [0.0 , 0.05) ┤▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 1
+   [0.05, 0.1 ) ┤ 0
+   [0.1 , 0.15) ┤▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 2
+                └                                        ┘
+                                Frequency
\ No newline at end of file
```

I think this is expected because the behavior of `master` wasn't quite right when computing the padding.
### `master`
```julia
    # this includes [rightmost edge, rightmost edge + binwidth] as part of the padding
    # even though this bin doesn't exist/get printed
    for I in eachindex(edges)
        val1 = float_round_log10(edges[I], binwidth)
        val2 = float_round_log10(val1+binwidth, binwidth)
        # ... compute padding
```
### this PR
```julia
    # only goes to [penultimate edge, rightmost edge]
    for i in 1:length(counts)
        binwidth = binwidths[i]
        val1 = float_round_log10(edges[i], binwidth)
        val2 = float_round_log10(val1+binwidth, binwidth)
        # ...
```
so I updated the reference for this test case.

I also added a test case for nonuniform binning.

johnnychen94 (approved):

Evizero (approved):
LGTM! thanks

PR #147, opened 2021-08-30
Switch to compact repr for floats

t-bltg:
I believe using a compact representation is more readable and eats less width.

```julia
using UnicodePlots
lineplot([cos, sin], -π/2, 2π, ylim=(-1.123456789, 1.123456789))
```

Before:
before

After:
after


johnnychen94 (approved):

t-bltg:
Way better looking now: http://docs.juliaplots.org/latest/generated/unicodeplots/

PR #148, opened 2021-08-30
Add trailing newline

t-bltg:
Consider the following example:

```bash
$ julia -e 'using UnicodePlots; show(lineplot([1, 2])); show(lineplot([3, 4]))'
```

**Without PR**
without

**With PR**
with

- [x] Generate new ref imgs

Not sure if it's the right way to fix this ...


t-bltg:
For now, fixed by https://github.com/JuliaPlots/Plots.jl/commit/4744152f86bd179e2fe5b6085715c23f857a5f36.

Issue #106, opened 2019-10-29
Handle NaNs and Infs in dataset gracefully


karlwessel:
Currently plotting something with NaNs or Infs like `f(x) = 1/x` fails with a rather unhelpful error message:

```
julia> using UnicodePlots

julia> f(x) = 1/x
f (generic function with 1 method)

julia> lineplot(f, 0,1)
ERROR: InexactError: trunc(Int64, -Inf)
Stacktrace:
 [1] trunc at ./float.jl:696 [inlined]
 [2] trunc at ./float.jl:353 [inlined]
 [3] floor at ./float.jl:357 [inlined]
 [4] ceil_neg_log10(::Float64) at .julia/packages/UnicodePlots/05CWJ/src/common.jl:38
 [5] round_up_subtick(::Float64, ::Float64) at .julia/packages/UnicodePlots/05CWJ/src/common.jl:42
 [6] plotting_range_narrow at /home/karl/.julia/packages/UnicodePlots/05CWJ/src/common.jl:56 [inlined]
 [7] extend_limits(::Array{Float64,1}, ::Tuple{Float64,Float64}) at .julia/packages/UnicodePlots/05CWJ/src/common.jl:71
 [8] #Plot#18(::String, ::String, ::String, ::Int64, ::Int64, ::Symbol, ::Tuple{Float64,Float64}, ::Tuple{Float64,Float64}, ::Int64, ::Int64, ::Bool, ::Bool, ::Type{Plot}, ::StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}, ::Array{Float64,1}, ::Type{BrailleCanvas}) at .julia/packages/UnicodePlots/05CWJ/src/plot.jl:133
 [9] (::getfield(Core, Symbol("#kw#Type")))(::NamedTuple{(:xlabel, :ylabel, :width),Tuple{String,String,Int64}}, ::Type{Plot}, ::StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}, ::Array{Float64,1}, ::Type{BrailleCanvas}) at ./none:0
[...]

PR #150, opened 2021-09-02
Replace !isnan with isfinite

t-bltg:
Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/106.

### DIFF
--- a/src/interface/lineplot.jl
+++ b/src/interface/lineplot.jl
@@ -101,8 +101,9 @@ function lineplot(
         color::UserColorType = :auto,
         name = "",
         kw...)
-    idx = map(!isnan, x)
-    idx .&= map(!isnan, y)
+    idx = map(x, y) do i, j
+        isfinite(i) && isfinite(j)
+    end
     new_plot = Plot(x[idx], y[idx], canvas; kw...)
     lineplot!(new_plot, x, y; color = color, name = name)
 end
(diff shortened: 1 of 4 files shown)


src/interface/lineplot.jl
```diff
@@ -101,8 +101,7 @@ function lineplot(
         color::UserColorType = :auto,
         name = "",
         kw...)
-    idx = map(!isnan, x)
-    idx .&= map(!isnan, y)
+    idx = map(isfinite, x) .& map(isfinite, y)
```
johnnychen94, line 104: This doesn't matter much but I'm sensitive to unnecessary memory allocations :)

```suggestion
    idx = map(x, y) do i,j
        isfinite(i) && isfinite(j)
    end
```

johnnychen94:
I hope we could always have a testset for every non-trivial PR so that we don't meet similar issues in the future.

src/interface/lineplot.jl
t-bltg: You're absolutely right !

t-bltg:
> I hope we could always have a testset for every non-trivial PR so that we don't meet similar issues in the future.

Will add, once I fix https://github.com/JuliaPlots/UnicodePlots.jl/pull/151, which turned to be non trivial to fix.

t-bltg:
Damn, the new map block now fails the CI :confused:

t-bltg:
Added a test in `tst_canvas.jl` and `tst_lineplot.jl`.

johnnychen94 (approved):

johnnychen94:
Evizero enjoys writing tests (and sometimes too verbose) so I think we should keep his good habit 😄

Issue #129, opened 2021-02-19
lineplot([cos,sin,tan],-2π,3π)


ixzh:
It didn't plot

t-bltg:
Offending loop:
https://github.com/JuliaPlots/UnicodePlots.jl/blob/65d3cd49072513f1d146ddea8bb9a6d31e70ff78/src/canvas.jl#L73

PR #151, opened 2021-09-02
Avoid drawing pixels out of canvas bounds

t-bltg:
The example at https://github.com/JuliaPlots/UnicodePlots.jl/issues/129 was drawing `> 1e17` pixels out of the canvas boundaries when `tan → ∞`.

This PR:
- adds a hard limit on number of points drawn (`typemax(Int32) ~ 𝓞(1e9)` pts per line);
- avoids drawing pixels out of canvas bounds.

Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/129.


src/canvas.jl
```diff
@@ -44,35 +44,35 @@ end
...
+    (y1 < my && y2 < my) || (y1 > My && y2 > My) && return c
+
+    x2p(x) = (x - mx) / width(c) * pixel_width(c)
+    y2p(y) = pixel_height(c) - (y - my) / height(c) * pixel_height(c)
+
+    Δx = x2p(x2) - (cur_x = x2p(x1))
+    Δy = y2p(y2) - (cur_y = y2p(y1))
+
+    nsteps = abs(Δx) > abs(Δy) ? abs(Δx) : abs(Δy)
+    nsteps = min(nsteps, typemax(Int32))  # hard limit
```
t-bltg, line 62: @johnnychen94, is this a reasonable fix ? :thinking:

src/canvas.jl
```diff
@@ -44,35 +44,35 @@ end
...
+    δy = Δy / nsteps
+
+    px, Px = extrema([x2p(mx), x2p(Mx)])
+    py, Py = extrema([y2p(my), y2p(My)])
+
     pixel!(c, floor(Int, cur_x), floor(Int, cur_y), color)
-    for i = 1:nsteps
-        cur_x += inc_x
-        cur_y += inc_y
+    for _ = 1:nsteps
```
johnnychen94, line 71: `1000` visually looks good to me.

```suggestion
    # For performance, we draw at most 1000 points for each line
    max_num_iter = 1000
    R = if nsteps > max_num_iter
        range(1, stop=nsteps, length=max_num_iter)
    else
        range(1, stop=nsteps, step=1)
    end
    for _ in R
```

This reduces the drawing time of `lineplot([cos, sin, tan], -π/2, 2π)` from 6.5s to 0.000228 seconds

src/canvas.jl
johnnychen94: LGTM but actually we could expect `nsteps` to be integers here since `for _ in 1:10.5` is equivalent to `for _ in 1:10`.

```suggestion
    nsteps = min(floor(Int, nsteps), typemax(Int32))  # hard limit
```

johnnychen94 (approved):
LGTM, with the performance comment resolved I think it's good to go.

src/canvas.jl
t-bltg: Sure, but what if someone wants to draw a line with more than `1000` pts ? I'd be cautious and leave a rather large upper bound ...

johnnychen94: I'm not sure... I would bet people won't be sensitive to more than 1000 pts for terminals... since we can't enlarge a terminal..

By calling `@show nsteps` with `lineplot([cos, sin, tan], -π/2, 2π)` I see normal `nsteps` are no-larger than 250.

src/canvas.jl
t-bltg: I remember now why I didn't cast nsteps to `Int`, it fails the CI.

src/canvas.jl
t-bltg: Unfortunately, limiting `nsteps` to even something as large as `100_000` fails the tests.

johnnychen94: Odd, in my local machine it only breaks the newly added `cos_sin_tan.txt`

johnnychen94: @t-bltg Anyway, since you're _de factor_ maintainer now I'll leave this to you to decide whether we want this "optimization". I will travel tomorrow so might not be available in the next few days.

t-bltg: By using `nsteps = min(nsteps, 100_000)  # hard limit`, and `pkg> test UnicodePlots` I get these errors:
```
Test Summary:     | Pass  Error  Total
tst_canvas.jl     |  116      5    121
  interface       |   84            84
  print and show  |   29      5     34
    BrailleCanvas |    4      1      5
    DensityCanvas |    4      1      5
    BlockCanvas   |    4      1      5
    AsciiCanvas   |    4      1      5
    DotCanvas     |    4      1      5
    HeatmapCanvas |    9             9
  color mixing    |    1             1
  colors          |    1             1
```

t-bltg: > since you're de factor maintainer
That's not a reason to do something messy :wink: 
I think I will leave that PR open until we find a better solution ...

johnnychen94: Ha! `range(1, stop=nsteps, length=100_000)` is different than `range(1, stop=min(nteps, 100_000))`.

t-bltg: > Ha! `range(1, stop=nsteps, length=100_000)` is different from `range(1, stop=min(nteps, 100_000))`.

End of the day :sleeping:

t-bltg: Ok, so I've limited the `nsteps` to `typemax(Int32)` and `max_num_iter` to `typemax(Int16)` for consistency:
```julia
@time lineplot([sin,tan],-2π,10π)
  0.001015 seconds (534 allocations: 56.203 KiB)
```
@johnnychen94, does it look acceptable ?

johnnychen94 (approved):

src/canvas.jl
t-bltg: Thanks for your help !

Issue #28, opened 2015-09-24
Implement a save() function to provide generic solution for plots


Evizero:
as discussed in  #19

Evizero:
@tbreloff I'll soon implement a save to txt file method. What would you suggest the method name/signature to be? `safefig(myplot, "filename.txt")` or would you suggest avoiding name collisions?

tbreloff:
Cool!  Shouldn't matter to me... I only ever import, so all calls are
prefaced (UnicodePlots.savefig, etc).  Feel free to use the same naming as
in Plots.

Evizero:
Alright. Looking at this now it seems I don't obey the subtle intended distinction between `show` and `print`, resulting in `print` writing color decorated text into any stream (even files), which make them unreadable. So that's the first thing to fix.

Other than that I am a bit reluctant now to make a special save method since it is probably best to just treat a `Plot` like one would a string (i.e. write them to a file the same way). An option to save the fully color decorated string may be useful though. especially if I ever want to create real tests for this package.

hustf:
Bump.
Perhaps there's an obvious way, without a modification here? If so, I haven't been able to figure it out. When print or show to a file, an IOStream I get:
	
"[37m"
and so forth.

I've been copying from the REPL to a text file, which works but gets boring.

digital-carver:
> Other than that I am a bit reluctant now to make a special save method since it is probably best to just treat a Plot like one would a string

In my opinion there's value in providing the common interface, such as the `savefig` method. A lot of the value of `Plots.jl` is in not having to relearn slightly different interfaces every time, instad having a common reliable interface. Having the "obvious thing" (saving with `savefig`) would lead a smoother interface, even if all it did underneath was a `print`.

crstnbr:
Kindly bumping again. Would love to have simple `savefig` function in UnicodePlots!

PR #152, opened 2021-09-02
Add basic `savefig` implementation

t-bltg:
Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/28.

### DIFF
--- a/src/UnicodePlots.jl
+++ b/src/UnicodePlots.jl
@@ -6,7 +6,6 @@ using StatsBase: Histogram, fit, percentile
 using SparseArrays: AbstractSparseMatrix, findnz
 
 export
-
     GraphicsArea,
         Canvas,
             BrailleCanvas,
@@ -50,7 +49,9 @@ export
     densityplot, densityplot!,
     heatmap,
     spy,
-    boxplot, boxplot!
+    boxplot, boxplot!,
+
+    savefig
 
 include("common.jl")
 
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -508,8 +508,9 @@ end
 """
     savefig(p, filename; color=false)
 
-Save a `Plot` to a file. The optional keyword argument `color` determines if the ANSI color codes
-should be included or not in the file.
+Print the given plot `p` to a text file.
+
+By default, it does not write ANSI color codes to the file. To enable this, set the keyword `color=true`.
 
 # Examples
 ```julia-repl
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -518,10 +518,12 @@ julia> savefig(lineplot([0, 1]), "foo.txt")
 
 ```
 """
-function savefig(p::Plot, fn::String; color::Bool=false)
-    ext = lowercase(splitext(fn)[2])
-    ext in (".png", ".jpg", ".jpeg", ".tif", ".gif", ".svg") && @warn "`UnicodePlots.savefig` only support writing to text files" filename=fn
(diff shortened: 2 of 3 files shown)


src/UnicodePlots.jl
```diff
@@ -50,7 +49,9 @@ export
     densityplot, densityplot!,
     heatmap,
     spy,
-    boxplot, boxplot!
+    boxplot, boxplot!,
+
+    savefig
```
t-bltg, line 54: Will this conflict with other packages ?

BeastyBlacksmith: Potentially, but I'd say we should strive for a common terminology rather than introduce slightly different terms with the same meaning to avoid conflicts

t-bltg: So you say it's OK to export savefig from `UnicodePlots` ?

BeastyBlacksmith: I'd say so.

src/plot.jl
```diff
@@ -504,3 +504,21 @@ function Base.show(io::IO, p::Plot)
         print_title(io, border_padding, p.xlabel, p_width = border_length)
     end
 end
+
+"""
+    savefig(p, fn[, color])
```
johnnychen94, line 509: We use `;` to indicate keyword argument, and with default values we meant it to be optional so no need to add extra `[]`.

Is `fn` a commonly used name in Plots? For me `filename` is clearer.

```suggestion
    savefig(p, filename; color=false)
```

src/plot.jl
```diff
@@ -504,3 +504,21 @@ function Base.show(io::IO, p::Plot)
...
+
+Save a `Plot` to a file. The optional keyword argument `color` determines if the ANSI color codes
+should be included or not in the file.
```
johnnychen94, line 512: How about this?

```suggestion
Print the given plot `p` to a text file.

By default, it does not write ANSI color codes to the file. To enable this, set the keyword `color=true`.
```

src/plot.jl
```diff
@@ -504,3 +504,21 @@ function Base.show(io::IO, p::Plot)
...
+
+# Examples
+```julia-repl
+julia> savefig(lineplot([0, 1]), "foo.txt")
+
+```
+"""
+function savefig(p::Plot, fn::String; color::Bool=false)
```
johnnychen94, line 520: I would imagine users writing `savefig(p, "out.png")`  and expect some png format results. Since we don't support it we should probably add a warning on some known extensions so that users don't get surprised when they try to open the file with their favorite image reader.

```suggestion
function savefig(p::Plot, fn::String; color::Bool=false)
    ext = lowercase(splitext(fn)[2])
    ext in (".png", ".jpg", ".jpeg", ".tif", ".gif", ".svg") && @warn "`UnicodePlots.savefig` only support writing to text files" filename=fn
```

johnnychen94 (approved):
If this integrates with Plots better then I have no objections.

FWIW, months ago I've made a libsixel wrapper https://github.com/johnnychen94/Sixel.jl and I did plan to experiment on gnuplots (see also https://github.com/csdvrx/sixel-gnuplot) and see if it's possible to bring sixel sequences to UnicodePlots, but then I got busy...

t-bltg:
> FWIW, months ago I've made a libsixel wrapper https://github.com/johnnychen94/Sixel.jl

I've learned a new word today, thanks :smiley:  !

Issue #9, opened 2015-08-19
Support for IJulia


diegozea:
¡Hi! At the moment, IJulia show the plots in the following way:

![ijulia](https://cloud.githubusercontent.com/assets/2822757/9357636/7943adfa-465e-11e5-8f82-10ecfeb963e5.png)

Evizero:
Not much I can do about that I think. The missing newline should be corrected in the current master, but the weird line stretching is due to the fact that the monospace font that your IJulia uses is not as monospace as it pretends to be

Evizero:
- [ ] but i will look into it a little closer. didn't really use IJulia yet

Evizero:
do the line-breaks now work for you? (don't forget Pkg.update())

Evizero:
Thank you for the report. This should be fixed now, so I am closing this issue.

The uneven line spacing issue in IJulia is due to the font that pretends to be monospace but really isn't. However, if you set `border=:dotted` it will look a lot less ugly.

![ijulia](https://cloud.githubusercontent.com/assets/10854026/9410470/becc8ad6-4820-11e5-86fd-ded88a850d82.png)

diegozea:
Great, thanks! :)

stevengj:
Regarding the font-spacing problem, can you give more detail?  It would be nice to file a bug report with Jupyter if this is the case.

(Of course, even in a monospace font, fonts and displays don't always agree about which glyphs are width 2 and which are width 1.)

stevengj:
Oh I see, the problem is the Braille characters?  For example, the exclamation marks don't line up for `println("a!\n⢀!\n")`

Evizero:
Right. it seems the braille character set is treated very differently from one font to the next

stevengj:
I filed a Jupyter bug.  Even if it is falling back to a non-monospaced font for the glyph of some Unicode character, it should still emulate a Terminal by using a layout grid based on the original monospace font.

Evizero:
thank you

t-bltg:
For reference: https://github.com/JuliaPlots/UnicodePlots.jl#know-issues.

PR #159, opened 2021-09-05
Update docs

t-bltg:
```bash
$ (cd doc; bash gen_imgs.sh)
```
, now generates most of the doc examples for `README.md`, to keep it up-to-date with the latest changes.

I've modified some examples to use `border = :dotted` as a workaround for https://github.com/JuliaPlots/UnicodePlots.jl/issues/9.


doc/gen_imgs.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+{
+  ver=2.x
+
+  ${JULIA-julia} gen_imgs.jl $ver
+
+  for f in imgs/$ver/*.txt; do
+    html=${f%.txt}.html
+    cat $f | ${ANSI2HTML-ansi2html} --input-encoding=utf-8 --output-encoding=utf-8 >$html
+    ${WKHTMLTOIMAGE-wkhtmltoimage} --quiet --crop-w 800 --quality 85 $html ${html%.html}.png
```
johnnychen94, line 10: Good to know this 👍

johnnychen94 (approved):
Nice!

At some point, we might reach large repo storage and make package downloading slower. As you may have noticed: some image files are permalinks to githubusercontent.com: people use this trick to not add the contents into git repos.

This PR will double the repo size. Another trick to work around this is to keep the shell scripts in `gh-pages` branch, keep the image data there, and use permalinks to target the images there. The Julia Pkg system only downloading contents from the commit and not the entire repo (say, `master` branch) so putting things to `gh-pages` won't make it slower for users to do `pkg> add UnicodePlots`

```console
$ du -sh
184K	doc
4.0K	LICENSE.md
4.0K	Manifest.toml
4.0K	Project.toml
 16K	README.md
212K	src
1.4M	test
```

This storage stuff isn't urgent at all so whether doing this trick totally depends on how you like it.

t-bltg:
> At some point, we might reach large repo storage and make package downloading slower. As you may have noticed: some image files are permalinks to githubusercontent.com: people use this trick to not add the contents into git history so as to get a thin repo.
> 
> This PR will double the repo size. Another trick to work around this is to keep the shell scripts in `gh-pages` branch, keep the image data there, and use permalinks to target the images there. The Julia Pkg system only downloading contents from the commit and not the entire repo (say, `master` branch) so putting things to `gh-pages` won't make it slower for users to do `pkg> add UnicodePlots`

I totally omitted the repo size when `DL`ing `UnicodePlots`, I only checked for the size of individual images displaying in the `README.md`.
Those are very relevant remarks !

This PR is only a workaround in some sense, I think this is cumbersome to use `Documenter` for the docs, and also, there are plot rendering issues as in `http://docs.juliaplots.org/latest/generated/unicodeplots/` which looks sparse between lines.

My preference to reduce the repo size will be to move the docs to another branch.
Is it mandatory to use `gh-pages`, I'm afraid we can accidentally remove those contents ?
Maybe we could use `unicodeplots-docs` ?

johnnychen94:
unicodeplots-docs sounds good to me. This isn't an urgent issue so really depends on if you want to future-proof this right now or fixes it in the future when someone complains it.

doc/gen_imgs.sh
t-bltg: This is hacky, but was the only sequence that allowed me to generate correct pngs.

`ansi2html` does a good job, but then the issue is to include `html` in a `markdown`, and I haven't found a way to do that, so I've defaulted to `png` conversion.

t-bltg:
> unicodeplots-docs sounds good to me. This isn't an urgent issue so really depends on if you want to future-proof this right now or fixes it in the future when someone complains it.

I'll do it right now, doubling the repo size is not acceptable for doc only (even if the initial repo size is small :wink:)

t-bltg:
Ok, it's done. I've added a note on how to update doc imgs: https://github.com/t-bltg/UnicodePlots.jl/tree/doc_imgs#documentation-update

PR #162, opened 2021-09-06
Bump version

t-bltg:
@johnnychen94, will all the potential breaks of the new commits, I think this should be `2.2.0` not `2.1.1`, right ?


johnnychen94 (approved):

johnnychen94:
According to semver, any new features should trigger a minor version bump.

PR #163, opened 2021-09-06
Log/ln scale consistency with `Plots.jl`

t-bltg:
While I was [integrating log scales](https://github.com/JuliaPlots/Plots.jl/pull/3796) in `Plots.jl`, I noticed an inconsistency with https://github.com/JuliaPlots/Plots.jl/blob/34758c82a34984a3627d7875ffdfa2e4a69da8a2/src/args.jl#L241.

Added `x/yscale` docs, will generate imgs after release.


johnnychen94:
Since we just trigger the release, so we can take it as a bug fix, maybe?

t-bltg:
> Since we just trigger the release, so we can take it as a bug fix, maybe?

Yes, as fast as possible, `2.2.1`.

PR #167, opened 2021-09-06
Fix incorrect braille canvas spacing

t-bltg:
In current `README.md` e.g. [scatterplot](https://github.com/JuliaPlots/UnicodePlots.jl#scatterplot) we observe incorrect xaxis label position.

It it due to using [space](https://unicode-table.com/en/0020/) (`U+0020`) symbol instead of [blank braille](https://unicode-table.com/en/2800/) (`U+2800`) symbol in `src/plots.jl`.

Before:
before

After:
after

- Breaking, because the x-axis labels now point inside the canvas, not on the border.
- All ref imgs generated, because we now pad blanks after each row to match the plot full width, also accounts for color bar.
- `Base.show` has been simplified.


johnnychen94:
Interesting, this is what I get from iTerm with [JuliaMono](https://github.com/cormullion/juliamono)

t-bltg:
> Interesting, this is what I get from iTerm with [JuliaMono](https://github.com/cormullion/juliamono)
> 
> 
> 
> I notice the canvas border (if this is the right term) is not aligned well in your "after" version, is that expected?

Sorry, I should have marked this as a WIP, still on it :wink:.

src/graphics.jl
```diff
@@ -7,20 +7,19 @@ function printrow end
 function Base.print(io::IO, c::GraphicsArea)
     for row in 1:nrows(c)
         printrow(io, c, row)
-        row < nrows(c) && print(io, '\n')
+        println(io)
```
johnnychen94, line 10: I believe it's deliberate to not add a new line at the end because Julia REPL will insert one. For example, with this PR there are two empty lines at the end.

johnnychen94 (approved):
LGTM with one comment on the suspicious empty lines.

src/plot.jl
```diff
@@ -354,31 +354,68 @@ function points!(plot::Plot{<:Canvas}, args...; vars...)
...
+        print(io, left_pad)
+        print_color(color, io, b[Symbol(loc, :l)], repeat(b[loc], length), b[Symbol(loc, :r)])
+        print(io, right_pad)
+    end
 end
 
 _nocolor_string(str) = replace(string(str), r"\e\[[0-9]+m" => "")
 
+function print_labels(io::IO, mloc::Symbol, p::Plot, border_length, left_pad::AbstractString, right_pad::AbstractString, blank::Char)
+    if p.show_labels
```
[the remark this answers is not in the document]
t-bltg: Changed.

src/graphics.jl
t-bltg: Fixed.

PR #171, opened 2021-09-09
Add support for text annotation on an existing plot

### LINKED ISSUE #170: [FR] Text annotations
Thanks for the great package! I love the idea of making quick plots and visualizations right in the REPL.

I was wondering if there was a way to make text annotations _within_ the plot (I see a lot of options for adding annotations around the boundary of the plot with the `annotate!` function). I think the answer is no right now, but I wanted to double check.

My use case is that I would like to make a lightweight REPL backend alternative for doing tensor network visualizations with my package [ITensorsVisualization.jl](https://github.com/ITensor/ITensorsVisualization.jl), which currently uses Makie as a backend so that the visualizations can be interactive. The graphs can be pretty complicated, in general tensor networks can be directed and weighted multigraphs with quite a bit of metadata (MetaMultiWeightedDiGraph?), but for the sake of simplicity for use with UnicodePlots it could be visualized as a simple undirected graph. However, in order to understand the graph that gets plotted, it would be very helpful to have text annotations to label the nodes and edges of the graph.

Here is a minimal example I cooked up making a network visualization using `GraphRecipes` with `UnicodePlots` as a backend:
```julia
using LightGraphs, GraphRecipes, Plots

n = 5
g = wheel_graph(n)

edgelabel_dict = Dict()
for i in 1:n, j in 1:n
  edgelabel_dict[(i, j)] = string("edge ", i, " to ", j)
end

gr()
plt_gr = graphplot(g, edgelabel=edgelabel_dict, names=vertices(g), markershape=:circle, curves=false)
savefig(plt_gr, "wheelgraph")

unicodeplots()
plt_up = graphplot(g, edgelabel=edgelabel_dict, names=vertices(g), markershape=:circle, curves=false)
```
outputs:
```julia
julia> plt_up
        ┌────────────────────────────────────────┐ 
    0.7 │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⡀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡠⠛⡟⢼⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⠔⠁⠀⢇⢸⠀⠉⠢⢄⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⢀⠀⠀⠀⠀⢀⠔⠑⢀⠀⠀⢸⢸⠀⢈⠐⠀⠈⠑⠢⣀⠀⠀⡁⠀⠀⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠠⠀⠀⢀⠔⠁⠀⠈⠠⠀⠀⠀⣿⠀⠀⠈⠀⠀⠀⠀⠀⠉⠒⠤⣀⠀⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠀⢀⠔⠁⠀⠀⠀⠀⠀⠀⠀⠀⢹⠀⠀⠀⠄⠀⠀⠀⠀⢀⣀⣀⣄⣑⢶⣶⠄⠀⠀⠀⠀│ 
        │⠤⠤⠤⠤⠤⠤⢤⠴⠥⠤⠤⣤⣤⣤⣤⡤⠤⠤⠤⢾⡷⠶⠶⠶⠭⠭⠭⠭⠥⠤⠤⢦⠴⠥⠤⠤⠤⠤⠤⠄│ 
        │⠀⠀⠀⠀⠐⠿⠯⢍⡙⠉⠉⠀⠀⠀⠀⠀⠐⠀⠀⢸⢣⠀⠀⠀⠀⠀⠀⠀⠀⢀⠔⠁⠀⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠀⠈⠒⠤⣀⠀⠀⠀⠀⠀⡀⠀⢸⠸⡀⠀⠀⠂⡀⠀⢀⠔⠁⠀⠀⠂⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠀⠀⢈⠀⠀⠉⠢⢄⡀⠀⠄⡁⢸⠀⡇⠀⠀⠁⢄⠔⠁⠀⠀⠀⠀⠁⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠑⠢⣀⢸⠀⢸⠀⢀⠔⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢹⠒⣬⣦⠊⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│ 
        │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠈⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│ 
   -0.7 │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│ 
        └────────────────────────────────────────┘ 
        ⠀-0.7⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀0.7⠀ 
```
The formatting is a bit off when pasting into Github, but I think you get the idea. Note that the labels don't show up on the nodes or edges. It appears to try to put some basic version of a box where the edge labels are supposed to go. Here is the output with a GR backend:

wheelgraph

I would be happy to help with adding this feature, but would need some pointers on how to get started. I tried starting by looking at how the `UnicodePlots.annotate!` function works but I couldn't tell how to generalize it to putting annotations at a custom location.

Some of this would no doubt need some coordination with the `GraphRecipes.jl` package and maybe even `Plots.jl`. For example, it would be nice if the `GraphRecipes` plot didn't show the border, labels, and grid by default for graph visualizations, but the `UnicodePlots` keyword arguments `(border=:none, labels=false, grid=false)` don't get passed along properly to `UnicodePlots` when I use `GraphRecipes.graphplots`. Those work fine if I circumvent `GraphRecipes` and make a custom graph visualization directly with `UnicodePlots.lineplot!`.

mtfishman:
Add support for text annotation on an existing plot. For example:
```julia
using UnicodePlots

plt = lineplot([1, 2, 7], [9, -6, 8], title = "My Lineplot");
annotate!(plt, 3, 5, "My text 1")
annotate!(plt, 5, -5, "My text 2")
annotate!(plt, 2, -6, "My text 3")
```
outputs: [annotate.txt](https://github.com/JuliaPlots/UnicodePlots.jl/files/7139154/example.txt)
<details>
<summary>Screenshot</summary>

</details>

Some TODO items and questions:

- [x] Add support for other `Canvas` backends (I've only added support for `BrailleCanvas` for now).
- [x] Decide on a name. I chose `annotate!` to match the [`Plots.jl` interface](http://docs.juliaplots.org/latest/generated/gr/#gr-ref20) but other suggestion like `text` or `note` were brought up in #170.
- [x] Make the point location more precise. Currently it uses a conversion from the specified point to the char coordinate in a bit of a rough way with a fixed rounding specification, but it looks like there is some more sophisticated rounding that is being done in functions like [`pixel!`](https://github.com/JuliaPlots/UnicodePlots.jl/blob/758daa5e669fd50b6318fcaa077f8fe4b117b724/src/canvas/braillecanvas.jl#L58-L75).
- [x] ~~Add more fine-grained control over the location of the text, such as `:center`, `:top`/`:bottom`, `:left`/`:right`, etc. Currently the first character of the text is roughly at the designated `x, y` coordinate on the plot. [Do we want this? Maybe it is needed for good compatibility with Plots?]~~ EDIT: In the latest, I added keyword arguments `halign ∈ (:left, :center, :right)` and `valign ∈(:top, :center, :bottom)` to control where the text goes around the point, following the convention of `Plots.annotate!`. The default is `:center` for both, so the text is centered at the location specified by the user.
- [x] ~~Currently it is using `:auto` color as default so gives a new automatically generated color for each new text annotation, is this the desired behavior?~~ EDIT: I've since changed the default color to `:normal`, which I think is better.
- [x] Right now, if all or some of the characters of the text annotation are outside of the character grid, they are ignored. Should they be automatically moved into the grid?
- [x] Add tests (at least for `BrailleCanvas`).
- [ ] Improve the documentation.

Closes #170.


johnnychen94:
@t-bltg For the CI, in the "Actions" of package(or organization) settings we can only require approval for new users

t-bltg:
Hi, I'm currently afk, I will review this on Monday 😉.

t-bltg:
> For the CI, in the "Actions" of package(or organization) settings we can only require approval for new users

Mmmh, it seems I don't have access to the repo (or organization) settings.

t-bltg:
> Add support for other Canvas backends (I've only added support for BrailleCanvas for now).

I'll try something today.

>  Decide on a name. I chose `annotate!` to match the [`Plots.jl` interface](http://docs.juliaplots.org/latest/generated/gr/#gr-ref20)

Yes, I think we should name it `annotate!` to match `Plots.jl`. On the other hand should we rename the existing `annotate!` (https://github.com/JuliaPlots/UnicodePlots.jl#methods) which adds a label at fixed positions (`top-left`, `bottom-center`, ...) ?

> Make the point location more precise. Currently it uses a conversion from the specified point to the char coordinate in a bit of a rough way with a fixed rounding specification, but it looks like there is some more sophisticated rounding that is being done in functions like pixel!.

We could move the rounding algorithm to a common function and use that in both methods.

> In the latest, I added keyword arguments halign ∈ (:left, :center, :right) and valign ∈(:top, :center, :bottom) to control where the text goes around the point, following the convention of Plots.annotate!. The default is :center for both, so the text is centered at the location specified by the user.

Very nice.

> I've since changed the default color to :normal, which I think is better.

Agreed.

> Right now, if all or some of the characters of the text annotation are outside of the character grid, they are ignored. Should they be automatically moved into the grid?

> Right now, if all or some of the characters of the text annotation are outside of the character grid, they are ignored. Should they be automatically moved into the grid?

If we can detect when it happens, we could ignore the characters and emit a warning to the user.

t-bltg:
Bugs:

```julia
using Unicodeplots
plt = lineplot([1, 2, 7], [9, -6, 8], title = "My Lineplot")
annotate!(plt, 7, 10, "My text 1", halign=:right)
annotate!(plt, 1, 10, "My text 2", halign=:left)
```
seem to be offseted by `1` towards the center of the canvas.

src/interface/annotate.jl
```diff
@@ -0,0 +1,60 @@
+
```
t-bltg, line 1: I think the content of this file should go into `src/plot.jl`

mtfishman:
Thanks for the fixes, tests, and other improvements! `label!` seems like a good name for the previous `annotate!` functionality.

As for the question about annotations that end up outside of the grid, I think a warning is unnecessary since the user should be able to see that it happened by looking at the resulting plot. I think it is analogous to adding a `lineplot!` to an existing plot: some of the line may get cut off if it runs outside of the current plot limits. So I'm ok with the current behavior.

I think the main thing left to do is add other Canvas backends. Thanks for offering to look into that, let me know if I can help in any way!

t-bltg:
> `label!` seems like a good name for the previous `annotate!` functionality.

That choice was arbitrary: we have the choice between [`decorate!`](https://github.com/JuliaPlots/UnicodePlots.jl/blob/b608635c74317023464f549e2d2d14b07fcf8e47/src/plot.jl#L314) and [`label!`](https://github.com/JuliaPlots/UnicodePlots.jl/blob/master/src/plot.jl#L301), to rename the `annotate!` @johnnychen94  what do you think is more appropriate ?

t-bltg:
@mtfishman I've added support for other canvases.

I've unchecked the `Make the point location more precise.` because it finally seems needed, for example:
```bash
$ cat test/references/plot/annotations_BrailleCanvas.txt
$ cat test/references/plot/annotations_AsciiCanvas.txt
```
, it seems we need a better positioning method. Could you have a look at it ?

t-bltg:
I've updated `:AsciiCanvas, :DotCanvas, :BlockCanvas` to use `UInt16` so that we can support annotations using a subset of `Unicode` (plane 0).

t-bltg:
Looks good now:
```bash
$ cat test/references/plot/annotations_DotCanvas.txt
      ┌────────────────────────────────────────┐ 
    1 │North West        North       North East│ 
      │                    :              ..'  │ 
      │                    :            .'     │ 
      │                    :         .''       │ 
      │                    :      ..'          │ 
      │                    :    .'             │ 
      │                    : .''               │ 
      │West.............Origin.............East│ 
      │                 .' :                   │ 
      │              .''   :                   │ 
      │           ..'      :                   │ 
      │         .'         :                   │ 
      │      .''           :                   │ 
      │   ..'              :                   │ 
   -1 │South West        South       South East│ 
      └────────────────────────────────────────┘ 
       -1                                     1  
```

I'm afraid annotations on a `BrailleCanvas` will never be positioned as wished for `test/references/plot/annotations_BrailleCanvas.txt` still doesn't look good.

mtfishman:
Looks great @t-bltg, thanks for fixing the positioning and adding support for the other backends! Anything else left to do?

t-bltg:
> Looks great @t-bltg, thanks for fixing the positioning and adding support for the other backends! Anything else left to do?

Thanks, I think it's ready (I've added a deprecation warning for the renaming of `annotate!`).
Let's maybe wait for @johnnychen94's review before merging.

t-bltg:
Maybe what's left is to improve the doc (`README.md`) to clarify `label!` and `annotate!`.

mtfishman:
Related to your latest commit, I was also wondering what is needed to make this feature available through Plots.jl. What would we need for that?

johnnychen94 (commented):
This is a nice improvement. Sorry, I'm afraid I don't have time to read these carefully so I'll defer to @t-bltg 's judgment.

Please also make sure we properly test the old `annotate!` usage still works; currently, it directly renames all deprecated `annotate!` to `label!` but we still want to test that `annotate!` works. These test for deprecated features can be added/moved to a `test/deprecated.jl` file

t-bltg:
Deprecated move to `test/tst_deprecated.jl`. I've restore the old `annotate!` methods surrounded with a `depwarn`.

I'm merging this now, we can always improve the docs later.

mtfishman:
@t-bltg thanks for your help on this! I'm looking forward to using this feature.

PR #172, opened 2021-09-14
Rename `xyscale` to `xyfact` in `heatmap`, `xyscale` as symbol in `histogram`

t-bltg:
- Fix conflict with `xyscale` (`:log10`, ...) from `Plots.jl`
- Use symbol default instead of function for transforms (`log10` -> `:log10`) in histogram.
- Some indent consistency.


src/graphics/bargraphics.jl
```diff
@@ -5,17 +5,21 @@ mutable struct BarplotGraphics{R<:Number} <: GraphicsArea
     max_freq::Number
     max_len::Int
     symbols::AbstractVector{String}
-    transform
+    transform::Function
```
Evizero, line 8: this seems breaking. What about a custom type that implements the call interface

t-bltg: Sorry about that if it's breaking. Can you provide a failing example ? I don't really see what you mean by "custom type that implements the call interface", can you provide a link to the julia docs about that maybe ?

Evizero: Honestly i don't remember specifics. I just seem to recall that duck typing was quite common and one could implement for custom types what should happen when you do `myobject(x)` where `myobject isa MyType`.

Its unlikely people are making use of it in this context so it should probably be fine. I just commented because it caught my eye

t-bltg: Ok, thanks. I will keep an eye open if any issue arises following this change.

johnnychen94: For instance, this is a typical functor, i.e., a type that implements the call interface.

```julia
struct Scale{T}
    r::T
end

(f::Scale)(x) = f.r * x
```

t-bltg: Thanks: the relevant documentation is here: https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects.

t-bltg: Fixed with https://github.com/JuliaPlots/UnicodePlots.jl/pull/178.

PR #179, opened 2021-09-17
Save space by moving labels

t-bltg:
When using `xlabel` and `ylabel`, use `label!` to save up space (especially width), when new keyword `compact=true` (default).
This is mostly useful when `min/max` strings are long (e.g. when using a scale), in conjunction with a long `ylabel`.
The height save is marginal (one line).

Example:
```julia
using UnicodePlots
x = y = collect(9:101);
lineplot(x, y, xlabel="x", ylabel="a very, very long label", xscale=:log10, yscale=:log10)
```

**before**
before

**after**
after

One can always restore the previous behavior by toggling the `compact=false` keyword:
```julia
lineplot(x, y, xlabel="x", ylabel="a very, very long label", xscale=:log10, yscale=:log10, compact=false)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,9 +221,14 @@ parameters.
 
     Plotting range for the y coordinate
 
-- `margin::Int = 3`:
+- `labels::Bool = true`:
 
-    Number of empty characters to the left of the whole plot.
+    Can be used to hide the labels by setting `labels=false`.
+
+  ```julia
+  lineplot(sin, 1:.5:20, labels=false)
+  ```
+    ![Labels Screenshot](https://github.com/JuliaPlots/UnicodePlots.jl/raw/unicodeplots-docs/doc/imgs/2.x/labels.png)
 
 - `border::Symbol = :solid`:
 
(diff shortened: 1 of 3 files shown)


johnnychen94 (commented):
Looks nice. Although in my understanding, making `compact=true` by default might count as breaking change and should trigger a major version bump. How does Plots handle this kind of UI change?

t-bltg:
> Looks nice. This isn't a strong opinion at all: in my understanding, making `compact=true` by default might count as breaking change and should trigger a major version bump. How does Plots handle this kind of UI change?

What we could do is leave `compact=false` in `UnicodePlots` and use `compact=true` in `Plots` where the concept of layouts (and multiple plots from left to right) exist. This PR integrates well with `Plots`.

johnnychen94:
This is a better (conservative) option from my point of view. But if UI changes aren't considered breaking, we can of course make default `compact=true`.

t-bltg:
Let's be conservative, I've updated it to use `compact=false` by default.

PR #181, opened 2021-09-24
Support markers

t-bltg:
Abuse the new `annotate!` function from https://github.com/JuliaPlots/UnicodePlots.jl/pull/171 to scatter data with arbitrary markers (`Symbol`, `Char` or `String`), although the precision is weaker than using `marker=:pixel` (for example when scattering a line).

Extends `scatterplot` to use vector of `marker`s and `color`s.
Added more unit-tests.

```julia
scatterplot(
    [1, 2, 3], [3, 4, 1],
    marker=[:circle, '😀', "∫"],
    color=[:red, nothing, :yellow],
    border=:dotted
)
```
scatter

Corresponding PR in `Plots.jl`: https://github.com/JuliaPlots/Plots.jl/pull/3845.

@mtfishman, this might be handy for your example at https://github.com/JuliaPlots/UnicodePlots.jl/issues/170#issue-991273649.


src/interface/scatterplot.jl
```diff
@@ -28,6 +28,8 @@ Arguments
 - **`name`** : Annotation of the current drawing to be displayed
   on the right
 
+- **`markers`** : choose from (:circle, :rect, :diamond, :hexagon, :cross, :xcross, :utriangle, :dtriangle, :rtriangle, :ltriangle, :pentagon, :star4, :star5, :star6, :star8, :vline, :hline, :+, :x) or use a `Char` or unit length `String`
```
johnnychen94, line 31:
```suggestion
**`marker`** : choose from $(keys(MARKERS)) or use a `Char` or unit length `String`
```

README.md
```diff
@@ -68,6 +68,14 @@ scatterplot(1:10, 1:10, xscale=:log10, yscale=:ln, border=:dotted)
 ```
 ![Scatterplot Screenshot2](https://github.com/JuliaPlots/UnicodePlots.jl/raw/unicodeplots-docs/doc/imgs/2.x/scatterplot2.png)
 
+`markers` is supported, choose from (:circle, :rect, :diamond, :hexagon, :cross, :xcross, :utriangle, :dtriangle, :rtriangle, :ltriangle, :pentagon, :star4, :star5, :star6, :star8, :vline, :hline, :+, :x) or use a `Char` or unit length `String`
```
johnnychen94, line 71:
```suggestion
`markers` is supported, choose from a `Char`, unit length `String`, or common symbol names such as `:circle` (more from `keys(UnicodePlots.MARKER)`)
```

src/interface/scatterplot.jl
t-bltg: Nice, I didn't know we could do that in a docstring.

t-bltg:
@johnnychen94, do you think we should also change `color` to `colors` and support a vector of colors here ?

johnnychen94:
> do you think we should also change color to colors and support a vector of colors here ?

This seems a natural need from my understanding. But TBH I'm don't know how it implies to UnicodePlots' current codebase.

t-bltg:
> > do you think we should also change color to colors and support a vector of colors here ?
> 
> This seems a natural need from my understanding. But TBH I'm don't know how it implies to UnicodePlots' current codebase.

Or on the contrary keep `color` and `marker` to the singular form.
Providing `markers = [:circle, '.']` seems more natural from a user POV, but `marker = [:circle, '.']` will be easier for us maintainers.
Uh, dilemma.

t-bltg:
I think I'm done here. Pinging @johnnychen94 for final review.

johnnychen94 (approved):
Nice!

PR #185, opened 2021-09-26
Force color in tests for `PkgEval`

t-bltg:
- Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/134.

```bash
$ julia --color=yes -e 'using Pkg; pkg"test UnicodePlots"'
$ julia --color=no  -e 'using Pkg; pkg"test UnicodePlots"'
```
, until https://github.com/KristofferC/Crayons.jl/issues/47 is fixed.

- Code cleanup, use `NamedTuple` instead of `Dict` (faster):

**before**
```julia
julia> using UnicodePlots
julia> @time lineplot([0, 4, 8], [10, 1, 10], color = :blue, name = "other line")
  1.096257 seconds (858.07 k allocations: 52.733 MiB, 4.19% gc time, 99.74% compilation time)
julia> @time lineplot([0, 4, 8], [10, 1, 10], color = :blue, name = "other line")
  0.000232 seconds (1.09 k allocations: 29.000 KiB)
```

**after**
```julia
julia> using UnicodePlots
@time lineplot([0, 4, 8], [10, 1, 10], color = :blue, name = "other line")
  0.922900 seconds (857.27 k allocations: 52.692 MiB, 1.76% gc time, 99.84% compilation time)
@time lineplot([0, 4, 8], [10, 1, 10], color = :blue, name = "other line")
  0.000172 seconds (1.09 k allocations: 29.000 KiB)
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -25,7 +25,7 @@ tests = [
 ]
 
 const RNG = StableRNG(1337)
-const T_SZ = (24, 80)
+const T_SZ = (24, 80)  # terminal size
 
 # see JuliaTesting/ReferenceTests.jl/pull/91
 test_ref(reference, actual) = @test_reference(reference, actual, render = BeforeAfterFull(), format = "TXT")
@@ -57,8 +57,10 @@ function stable_sprand(r::AbstractRNG, m::Integer, n::Integer, density::Abstract
   sparse(I, J, V)
 end
 
-for test in tests
-    @testset "$test" begin
-        include(test)
+withenv("FORCE_COLOR"=>"X") do  # github.com/JuliaPlots/UnicodePlots.jl/issues/134
+    for test in tests
+        @testset "$test" begin
+            include(test)
+        end
     end
 end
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -57,8 +57,10 @@ function stable_sprand(r::AbstractRNG, m::Integer, n::Integer, density::Abstract
   sparse(I, J, V)
 end
 
-for test in tests
-    @testset "$test" begin
-        include(test)
+withenv("FORCE_COLOR"=>"X") do  # github.com/JuliaPlots/UnicodePlots.jl/issues/134
```
johnnychen94, line 60: Didn't know this is possible, thanks for the fix!

johnnychen94 (approved):

test/runtests.jl
t-bltg: That's a `Crayons` feature:
https://github.com/KristofferC/Crayons.jl/blob/2619561156a02fdf739c8146e09a96ceb176f43c/src/crayon.jl#L1

PR #186, opened 2021-10-01
Improve coverage, add tests

t-bltg:
- Improve coverage (hopefully to `100%`)
- ~~Add some `@inbounds` for `for` loops~~
- Add more `@warn` tests, uncovered cases
- Move `compact` closer to `ctor`


johnnychen94 (commented):
It is not guaranteed that all vectors are 1-based indexing.

```julia
using OffsetArrays
v = OffsetVector([1, 2, 3], -1)
v[0] == 1 # true
```

so if we want to add `@inbounds` for this method, we also need to ensure that `axes(X) == axes(Y) == axes(color)`, not just comparing their lengths, but also their offsets.

There is another way to ensure this, by offsetting all arrays into 1-based indexing, for example, `OffsetArrays.no_offset_view(x)` is guaranteed to be 1-based indexing for common array types.

t-bltg:
> It is not guaranteed that all vectors are 1-based indexing.
> 
> ```julia
> using OffsetArrays
> v = OffsetVector([1, 2, 3], -1)
> v[0] == 1 # true
> ```
> 
> so if we want to add `@inbounds` for this method, we also need to ensure that `axes(X) == axes(Y) == axes(color)`, not just comparing their lengths, but also their offsets.
> 
> There is another way to ensure this, by offsetting all arrays into 1-based indexing, for example, `OffsetArrays.no_offset_view(x)` is guaranteed to be 1-based indexing for common array types.

Ok, I'm gonna take the lazy path and remove `@inbounds`.

johnnychen94 (approved):

PR #190, opened 2021-11-28
Axis scale: change default exponent to unicode

t-bltg:
Proposition for https://github.com/JuliaPlots/Plots.jl/issues/3959.

Changes the default exponent to using unicode superscripts (https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts) e.g. `10¹⸳⁵`.

There is a discussion on which character to choose for the decimal separator, if any, since there is no consensus (see https://stackoverflow.com/questions/34350441/is-there-an-unicode-symbol-for-superscript-comma).
See also: https://en.wikipedia.org/wiki/Interpunct#Similar_symbols.

Rendering depends on the font used, but with `JuliaMono`, `scatterplot(1:10, 1:10, xscale=:log10, yscale=:ln, border=:dotted)` gives:
unicode_exp

One can always revert to using the old ascii format (e.g. `10^1.5`) using the added keyword `unicode_exponent=false`,
`scatterplot(1:10, 1:10, xscale=:log10, yscale=:ln, border=:dotted, unicode_exponent=false)` gives:
ascii_exponent

- [x] add tests
- [x] new ref imgs
- [x] update docs
- [x] new doc imgs (separate PR)


t-bltg:
Investigating separator: my choice would go for `U+2E31: Word Separator Middle Dot`:
```julia
for c ∈ ('·', 'ˑ', '·', 'ּ', '᛫', '•', '‧', '∘', '∙', '⋅', '⏺', '●', '◦', '⚫', '⦁', '⸰', '⸱', '⸳', '・', 'ꞏ', '･', '𐄁')
  println("\n\t10¹$(c)²³⁴⁵⁶    " * repr("text/plain", c))
end
```

johnnychen94 (approved):
Thanks for your consistent improvement to this package ❤️

PR #193, opened 2021-12-14
Spy: add zeros pattern

t-bltg:
1. Allow showing zeros pattern instead of non-zeros;
2. Remove stored zeros from shown pixels (previously included in `< 0`), see below example which is now correct;
3. Add default label when `color != :auto`;
4. Allow passing `xlabel` to `spy(...)`, which was previously overriden;

```julia
julia> using SparseArrays, UnicodePlots
julia> A = sparse([1, 1, 2, 3], [1, 3, 2, 3], [0, 1, 2, 0])
3×3 SparseMatrixCSC{Int64, Int64} with 4 stored entries:
 0  ⋅  1
 ⋅  2  ⋅
 ⋅  ⋅  0
julia> spy(A, border=:dotted)
     Sparsity Pattern    
     ⡤⠤⠤⠤⠤⠤⢤    
   1 ⡇⠀⠀⠀⠠⠀⢸ > 0
   3 ⡇⠀⠀⠈⠀⠀⢸ < 0
     ⠓⠒⠒⠒⠒⠒⠚    
     ⠀1⠀⠀⠀3⠀    
     2 nonzeros 
julia> spy(A, show_zeros=true, border=:dotted)
     Sparsity Pattern    
     ⡤⠤⠤⠤⠤⠤⢤    
   1 ⡇⠀⠄⠠⠀⠀⢸ = 0
   3 ⡇⠀⠅⠠⠨⠀⢸    
     ⠓⠒⠒⠒⠒⠒⠚    
     ⠀1⠀⠀⠀3⠀    
     7 zeros
```

Another example, `stdlib` shows all stored entries (zeros included) in the braille pattern: 
```julia
julia> using SparseArrays, UnicodePlots
julia> A = sparse([1, 1, 2, 3, 16], [1, 3, 2, 3, 16], [0, 1, 2, 0, 1])
16×16 SparseMatrixCSC{Int64, Int64} with 5 stored entries:
⠑⠅⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⢀
julia> spy(A, border=:dotted)
      Sparsity Pattern    
      ⡤⠤⠤⠤⠤⠤⠤⠤⠤⢤    
    1 ⡇⠐⠁⠀⠀⠀⠀⠀⠀⢸ > 0
      ⡇⠀⠀⠀⠀⠀⠀⠀⠀⢸ < 0
      ⡇⠀⠀⠀⠀⠀⠀⠀⠀⢸    
   16 ⡇⠀⠀⠀⠀⠀⠀⠀⢀⢸    
      ⠓⠒⠒⠒⠒⠒⠒⠒⠒⠚    
      ⠀1⠀⠀⠀⠀⠀16⠀    
      3 nonzeros
julia> spy(A, show_zeros=true, border=:dotted)
      Sparsity Pattern    
      ⡤⠤⠤⠤⠤⠤⠤⠤⠤⢤    
    1 ⡇⣯⣾⣿⣿⣿⣿⣿⣿⢸ = 0
      ⡇⣿⣿⣿⣿⣿⣿⣿⣿⢸    
      ⡇⣿⣿⣿⣿⣿⣿⣿⣿⢸    
   16 ⡇⣿⣿⣿⣿⣿⣿⣿⡿⢸    
      ⠓⠒⠒⠒⠒⠒⠒⠒⠒⠚    
      ⠀1⠀⠀⠀⠀⠀16⠀    
      ⠀253 zeros
```


src/interface/spy.jl
```diff
@@ -100,7 +102,23 @@ See also
 [`BrailleCanvas`](@ref), [`BlockCanvas`](@ref),
 [`AsciiCanvas`](@ref), [`DotCanvas`](@ref)
 """
-spy(A::AbstractMatrix; kwargs...) = spy(size(A)..., _findnz(A)...; kwargs...)
+
+function spy(A::AbstractMatrix; kwargs...)
+    rows, cols, vals = _strict_non_zeros(_findnz(A)...)
+    if get(kwargs, :zeros, false)
+        I = CartesianIndex.(zip(rows, cols))  # non zeros
```
johnnychen94, line 109: How about

```julia
rows, cols, vals = _findnz(A)
nonzero_R= findall(!iszero, vals)
if get(kwargs, :zeros, false)
    I = [CartesianIndex(rows[idx], cols[idx]) for idx in nonzero_R]
    mask = trues(size(A))
    mask[I] .= false;
    Z = CartesianIndices(axes(A))[mask]
    ...
end
```

t-bltg: Damn I really feel stupid on this one :sweat_smile:, thanks !

johnnychen94: The idea of `I = [CartesianIndex(rows[idx], cols[idx]) for idx in nonzero_R]` is that you don't need to call `rows[I]` to allocate two memory blocks for `Array{Int}`; thus it can saves two memory allocations.

src/interface/spy.jl
```diff
@@ -56,6 +56,8 @@ Arguments
 
 - **`canvas`** : The type of canvas that should be used for drawing.
 
+- **`zeros`** : Show zeros pattern instead of nonzeros (default).
```
johnnychen94, line 59: Just a confirmation; is this still the default? My understanding is that "by default it is `false`"?

I don't know if this matters if taking Plots.jl into account, but maybe `show_zeros` is easier to understand?

johnnychen94 (approved):
I don't know the details of this `spy` function (never used it) but generally looks good to me.

src/interface/spy.jl
t-bltg: You still need `rows[I]` and `cols[I]` with zeros eliminated in the call to `spy(size(A)..., rows, cols, vals; kwargs...)`, so I don't think we can avoid new allocation here ?

src/interface/spy.jl
t-bltg: The default of `spy(A)` is to show the sparsity pattern by coloring nonzeros entries (https://www.mathworks.com/help/matlab/ref/spy.html) in a sparse matrix (very useful in the context of finite differences, fem, applying boundary conditions, ...). So yes showing zeros is by default false. Agreed, another argument for renaming to `show_zeros` is that is does not shadow the builtin `zeros` from julia's stdlib. Showing zeros for a sparse matrix instead of nonzeros is not very used, but I find it occasionally useful !

Issue #153, opened 2021-09-02
[FR] Support sixel encoding


johnnychen94:
It would be super cool if we can bring sixel support for terminal plots, I've made a libsixel wrapper https://github.com/johnnychen94/Sixel.jl Check https://github.com/csdvrx/sixel-gnuplot for the sixel gallery.

There are two possible solutions:

* easy but slow one: for any plot library that can output an image or `Array{<:Colorant}`, pass the colorant array to `Sixel.sixel_encode` and thus get a nice terminal view
* difficult but (maybe) efficient one: https://en.wikipedia.org/wiki/ReGIS I'm not sure if Sixel.jl is ready for ReGIS format.

johnnychen94:
Looks like GR also have built-in sixel support https://discourse.julialang.org/t/plots-in-the-terminal-with-sixel/22503/4

![image](https://user-images.githubusercontent.com/8684355/131879525-020b2169-f825-465f-87e5-95cc8adecb35.png)

johnnychen94:
@t-bltg Will it be too tricky for Plots to add a virtual backend for Sixel by 1) loading the PNG output of another backend, 2) encode to sixel using Sixel.jl, and 3) print to terminal? Maybe we can pirate the `UnicodePlotsBackend` when sixel is supported by the terminal?

https://github.com/JuliaImages/ImageInTerminal.jl/pull/50 is what I did for ImageInTerminal to auto-switch to sixel encoding.

t-bltg:
I'll give it a try in the next few days !

waldyrious:
Just a side note that supporting sixel would probably make this package more like a TerminalPlots.jl than a UnicodePlots.jl :)

t-bltg:
@johnnychen94 I think we can work on this now.

I don't think I fully understand the sixel format, could you point to we a simple implementation (what's the base type to encode a sixel ?).

johnnychen94:
You can think of sixel as an advanced ANSI sequence for image types.

```julia
using Sixel
img = testimage("mandril")
sixel_encode(img)
```

t-bltg:
How can I determine how many sixel there are for this image of `512×512` RGB values ?.

johnnychen94:
The technical format says that one control code consists of 6 pixels, which is why the term "sixel" comes from ("six pixels"). Given that Sixel.jl is currently only a wrapper package for libsixel, there isn't a way to do pixel-level encoding.

I was thinking of something like

[...]

PR #196, opened 2021-12-18
add `imageplot` (support sixel)

t-bltg:
Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/153, goes with https://github.com/JuliaPlots/Plots.jl/pull/4011.

**with `Sixel` [^1]**
[^1]: for sixel support, code snippet must be ran under a [compatible terminal](https://github.com/johnnychen94/Sixel.jl#terminals-that-support-sixel).
```julia
using Sixel, UnicodePlots, ImageInTerminal, TestImages, ImageTransformations

main() = begin
  img = imresize(testimage("mandril_color"), ratio=.4)
  display(UnicodePlots.image(img, title="mandril sixel"))
  return
end

main()
```
mandril

**without `Sixel`**
```julia
using UnicodePlots, ImageInTerminal, TestImages, ImageTransformations

main() = begin
  img = imresize(testimage("mandril_color"), ratio=.4)
  ImageInTerminal.use_24bit()
  display(UnicodePlots.image(img, title="mandril 24 bit"))
  ImageInTerminal.use_256()
  display(UnicodePlots.image(img, title="mandril 256 color"))
  return
end

main()
```
mandril_24_256

Discussion:
- ~~naming ?, should we export this ?~~ chose to unexport `image`
-  ~~img is downsized using `restrict` in `ImageInTerminal`, is there a solution to get the ratio so that we can add a centered title and/or labels ?~~ solved using [terminal control sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html) to query the cursor carret size
- is the approach using `@require` standard, for optional dependency `ImageInTerminal`, or should we instead add `ImageInTerminal` as a hard dependency ?
- the sixel implementation was tricky, had to use some ansi escape codes for printing the right border, maybe we can do something better (see the following mwe for writing after a sixel sequence)

```bash
$ printf '\ePq
#0;2;0;0;0#1;2;100;100;0#2;2;0;100;0
#1~~@@vv@@~~@@~~$
#2??}}GG}}??}}??-
#1!14@\e\\\e[A\e[2Cworld
'
```
hi


johnnychen94:
I hope we can pend this PR for a while until we address https://github.com/JuliaImages/ImageInTerminal.jl/issues/48. The current ImageInTerminal has one unwanted side effect; that any package loads ImageInTerminal would change how colorant arrays are rendered because [it overrides `Base.show`](https://github.com/JuliaImages/ImageInTerminal.jl/blob/4cd5bebf718196cb49a66b600b27d4023696ead4/src/ImageInTerminal.jl#L63-L73).

t-bltg:
> I hope we can pend this PR for a while until we address [JuliaImages/ImageInTerminal.jl#48](https://github.com/JuliaImages/ImageInTerminal.jl/issues/48).

Thanks, for pointing that. Maybe I can help on that matter ?

The current ImageInTerminal has one unwanted side effect; that any package loads ImageInTerminal would change how colorant arrays are rendered in terminal even if you don't call it with UnicodePlots.image because it overrides Base.show.

Is this something we want to change for `ImageInTerminal` 1.0 ?

johnnychen94:
> Is this something we want to change for ImageInTerminal 1.0 ?

Yes, I'd like to separate ImageInTerminal into two packages: backends and frontends. The backend might be as simple as `ascii_encode(img)` and the frontend would provide a Display type and override `Base.display` and `Base.show` method. For instance, Sixel.jl can be seen as an ImageInTerminal backend.

To do this, we need to rework and deprecate the current `imshow` implementation: 1) introduces a pure encoding backend `ascii_encode([io], img, method; kwargs...)` as I did in Sixel.jl, make it a standalone package (say `AsciiImageEncoding`, a bad package name, but you can get the point), and 2) introduce a `TerminalGraphicDisplay` type and keep the previous `Base.show` methods in ImageInTerminal.

By doing this, this PR can directly use the `ascii_encode` methods in the new backend package just like using `sixel_encode` in Sixel. We could also move some core encoding functionality of the UnicodePlots to the new backend package.

johnnychen94:
> Thanks, for pointing that. Maybe I can help on that matter ?

Yeah, given that I have too many items on my todo-list it would be great if you can help implement this! I just sent you an invitation to ImageInTerminal, you can choose to implement the backend package in it as a sub-package like [SnoopCompileCore](https://github.com/timholy/SnoopCompile.jl), or you can create a new package and refactor the encoding part of  ImageInTerminal to the new package.

t-bltg:
Thanks, I'll continue the discussion in https://github.com/JuliaImages/ImageInTerminal.jl/issues/52 about details.

Project.toml
```diff
@@ -25,10 +25,10 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
-ColorSchemes = "3.19"
+ColorSchemes = "≥ 3.19"
```
johnnychen94, line 28: It's not a good idea to allow arbitrary breaking releases with `>=`. You can't ensure this package works with ColorSchemes v4, or Crayons v5 unless they're released.

A lot of Julia packages used this, and then the entire ecosystem shifts to the more restricted version. https://github.com/JuliaRegistries/General/pull/11114

Project.toml
```diff
@@ -18,34 +18,38 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
...
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
-ColorSchemes = "3.19"
+ColorSchemes = "≥ 3.19"
```
johnnychen94, line 28: It's not a good idea to allow arbitrary breaking releases with `>=`. You can't ensure this package works with ColorSchemes v4, or Crayons v5 unless they're released.

A lot of Julia packages used this, and then the entire ecosystem shifts to the more restricted version. https://github.com/JuliaRegistries/General/pull/11114

test/tst_imageplot.jl
```diff
@@ -0,0 +1,21 @@
+if isinteractive()  # FIXME: Sixel.query_terminal(...) fails when !(stdin isa TTY)
+    img = testimage("monarch_color_256")
+
+    @testset "sixel" begin
+        old_enc = ImageInTerminal.ENCODER_BACKEND[]
+        ImageInTerminal.ENCODER_BACKEND[] = :Sixel
+        p = imageplot(img, title = "sixel")
+        @test p.graphics.sixel[]
+        test_ref("imageplot/img_sixel.txt", @show_col(p, :displaysize => T_SZ))
```
johnnychen94, line 9: It seems that this file is not commited.

t-bltg: Xref https://github.com/JuliaIO/Sixel.jl/pull/18#issuecomment-1231525862.

Project.toml
t-bltg: Do you say that all the dependencies like `StaticArrays`, `Unitful`, ... currently at `1` should be pinned to a minor range instead ?

johnnychen94: > https://pkgdocs.julialang.org/v1/compatibility/#Version-specifier-format
> Caret specifiers are the default and hence 1.2.3 == ^1.2.3. 

I was speaking specifically about the `>=` usage (can't mind reading and realize that `>=` is a typo). Caret is okay, but `^3.19` equivalent to `3.19`.

test/tst_imageplot.jl
johnnychen94: But if this file is not commited, how would you expect other people to run the test in interactive mode?

Project.toml
t-bltg: I will revert to caret, sorry.  `^3.19` means `[3.19.0, 4.0.0)` according to https://pkgdocs.julialang.org/v1/compatibility/#Caret-specifiers, which is what I originally meant.

test/tst_imageplot.jl
t-bltg: Yeah, this is a wip since I get different results when running tests in `mlterm ` or `xterm`.

t-bltg: I don't know what to do here, should I remove the tests, since non-interactive mod is bound to fail ?

t-bltg:
I will probably merge this in the next few days, unless objection(s).

Issue #169, opened 2021-09-06
[FR] stand-alone version of `printcolorbarrow` to draw a colorbar in isolation


SimonEnsemble:
this is a feature request to draw a colorbar in isolation with a specific range, via something like
```julia
draw_colorbar(colormap = [(x.r, x.g, x.b) for x in ColorSchemes.leonardo.colors], title="concentration", range=(0.0, 10.0))
```
to give

![image](https://user-images.githubusercontent.com/30643894/132265694-466df1db-df65-4855-bc0f-02cf436696fd.png)

instead of using this:
```julia
using UnicodePlots, ColorSchemes
heatmap(10*rand(50, 50); title = "Hot!",
           colormap = [(x.r, x.g, x.b) for x in ColorSchemes.leonardo.colors])
```
(basically, I want to remove the un-needed heatmap from this portion of the code).

thx, awesome package!

PS I need this to go along with `GraphPlot` to show the colormap that colors nodes in a graph.

t-bltg:
@SimonEnsemble, sorry for the delay, this was not on top of my priority list.

Using https://github.com/JuliaPlots/UnicodePlots.jl/pull/199, is the syntax for plotting an isolated colorbar `Plot([0], [0], colorbar=true, colormap=:cividis, width=0, min_width=0)` ok for you ?

SimonEnsemble:
it works, thank you!

<img width="151" alt="image" src="https://user-images.githubusercontent.com/30643894/149817138-e5c07aa4-d15f-4dc2-a8e8-d3027a7c15bb.png">

though would be nice to somehow make `plot([], [], ...)` work to remove the bounding box/plot so that it is *just* a colorbar.

t-bltg:
Yes, this would be done it a following PR (which I'm writing now, since https://github.com/JuliaPlots/UnicodePlots.jl/pull/199 has been split in multiple PRs.

t-bltg:
PR is now open at https://github.com/JuliaPlots/UnicodePlots.jl/pull/205.

![isolated](https://user-images.githubusercontent.com/13423344/149818250-bfd47c33-b68d-4e26-845f-4e27ff426a8f.png)

SimonEnsemble:
😍 nice! it works for me!
is there an option to change the colorbar limits too? this is stuck at 0-1.
this will be useful for when coloring e.g. nodes of graphs according to some scheme.

t-bltg:
[...]

PR #199, opened 2022-01-12
Add `contourplot`

t-bltg:
- add `contourplot` support (`levels`, `colormap`, `colorbar` controls)
- add `Contour.jl` dependency for computing levels using the [`Marching_squares`](https://en.wikipedia.org/wiki/Marching_squares) algorithm
- add `blend` flag on all canvases to plot with/without color blend (avoid overlap issues on contour lines)
- add `visible` flag on all canvases: fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/169 as a side effect
- add more cmaps: [`cividis`](https://arxiv.org/ftp/arxiv/papers/1712/1712.01662.pdf), and some ANSI `256` subset colormaps.

https://github.com/JuliaGeometry/Contour.jl is a small requirement, since it only depends on `StaticArrays`.

See examples at https://github.com/JuliaPlots/Plots.jl/pull/4031.

The new example from `README.md` now looks like this (gaussian profile):
contourplot1

**tiny benchmark**
```julia
using UnicodePlots, BenchmarkTools

main() = begin
  x = -3:.1:3
  y = -7:.1:3
  X = repeat(x', length(y), 1)
  Y = repeat(y, 1, length(x))
  g(x, y, x₀=0, y₀=-2, σx=1, σy=2) = exp(-((x - x₀) / 2σx)^2 - ((y - y₀) / 2σy)^2)
  z = map(g, X, Y)
  @btime heatmap($z)  #   15.090 ms (197665 allocations: 9.45 MiB)
  @btime contourplot($x, $y, $z)  #   2.833 ms (29227 allocations: 1.18 MiB)
end

main()
```


src/canvas/densitycanvas.jl
```diff
@@ -1,4 +1,4 @@
-const den_signs = Ref([" ", "░", "▒", "▓", "█"])
+const den_signs = Ref([' ', '░', '▒', '▓', '█'])
```
johnnychen94, line 1: This change should be committed in a separate PR.

src/canvas.jl
```diff
@@ -229,3 +229,52 @@ function annotate!(
...
+        print(io, plot_padding)
+        # print z label
+        if row == div(nrows(c), 2) + 1
+            label = zlabel
+            print(io, label)            
+        end
+    end
+    print(io, repeat(blank, max_len - length(label)))
+    nothing
+end
```
johnnychen94, line 279: This code movement from `src/canvas/heatmapcanvas.jl` should be reverted.

src/colormaps.jl
```diff
@@ -14,1059 +14,1404 @@
...
-               [0.989815, 0.934329, 0.690198],
-               [0.989434, 0.941470, 0.697519],
-               [0.989077, 0.948604, 0.704863],
-               [0.988717, 0.955742, 0.712242],
-               [0.988367, 0.962878, 0.719649],
-               [0.988033, 0.970012, 0.727077],
-               [0.987691, 0.977154, 0.734536],
-               [0.987387, 0.984288, 0.742002],
-               [0.987053, 0.991438, 0.749504]]
+const _magma_data = [(0.001462, 0.000466, 0.013866),
```
johnnychen94: This should also be committed in a separate PR.

src/common.jl
```diff
@@ -133,36 +133,36 @@ function superscript(s::AbstractString)
...
         ma = mi + 1
         mi = mi - 1
     end
-    if string(scale) != "identity"
-        return fscale(mi, scale), fscale(ma, scale)
+    return if string(scale) != "identity"
+        fscale(mi, scale), fscale(ma, scale)
     else
-        return all(iszero.(limits)) ? plotting_range_narrow(mi, ma) : (mi, ma)
+        all(iszero.(limits)) ? plotting_range_narrow(mi, ma) : (mi, ma)
```
johnnychen94, line 165: unnecessary style changes to me.

src/common.jl
```diff
@@ -307,8 +307,8 @@ julia_color(color::Nothing)::JuliaColorType = :normal
 julia_color(color::Symbol)::JuliaColorType = color
 julia_color(color)::JuliaColorType = julia_color(crayon_256_color(color))
 
-@inline function set_color!(colors::Array{ColorType,2}, x::Int, y::Int, color::ColorType; force::Bool=false)
-    if color === nothing || colors[x, y] === nothing || force
+@inline function set_color!(colors::Array{ColorType,2}, x::Int, y::Int, color::ColorType, blend::Bool)
```
johnnychen94, line 310: This change can be breaking as it deletes the `force` keyword.

src/graphics/boxgraphics.jl
```diff
@@ -76,13 +79,13 @@ function printrow(io::IO, c::BoxplotGraphics, row::Int)
...
-    median_char = ['┬', '│' , '┴'][series_row]
-    right_box_char = ['┐', '├' , '┘'][series_row]
-    max_char = ['╷', '┤' , '╵'][series_row]
+    min_char = ('╷', '├' , '╵')[series_row]
+    line_char = (' ', '─' , ' ')[series_row]
+    left_box_char = ('┌', '┤' , '└')[series_row]
+    line_box_char = ('─', ' ' , '─')[series_row]
+    median_char = ('┬', '│' , '┴')[series_row]
+    right_box_char = ('┐', '├' , '┘')[series_row]
+    max_char = ('╷', '┤' , '╵')[series_row]
```
johnnychen94, line 88: again, should be submitted in a separate PR.

johnnychen94 (commented):
It's almost impossible to review a 1000-lines PR with arbitrary unorganized commits and changes so I only left a few comments on changes I think should be reverted or separated. There are still many out there that I didn't comment.

It would be clearer to separate the `visible` and `blend` changes in separate PR(s), and then make this `countourplot` PR an incremental feature on top of those PRs.

FWIW, for a clean git history: it makes no sense to record how you evolve into the current status, instead, it makes sense to record how functionalities are built step by step. Thus if you add something in the first commit and then immediately delete it in the second commit, it should be purged from the git history.

src/canvas.jl
t-bltg: Since it is now used in `contourplot` (which uses an arbitrary canvas) I think it should be moved away from `heatmapcanvas.jl` and put in `canvas.jl`. But I think this can be done in a separate colormap PR as you suggested.

src/common.jl
t-bltg: This isn't exposed in `README.md` (what I would call our API), so is this technically breaking ? This keyword is more intuitive since we do color mixing / blending after all.

t-bltg:
Thanks for taking the time to reviewing this. The workflow is difficult since on this branch, I needed some colormap changes, and the integration / testing within Plots.jl is best if I apply the changes in a single branch, but I hear the concern on reviewing my PRs. Developing using multiple branches is also hard to follow, and wastes development time. Sorry about the code style changes, bad practice indeed. I will rebase the whole thing and split PRs.

t-bltg:
Superseded by https://github.com/JuliaPlots/UnicodePlots.jl/pull/204.

PR #201, opened 2022-01-13
Fix square matrices (`heatmap` / `spy`)

t-bltg:
Fix for displaying square matrices on `heatmap` and `spy`. [This comment](https://github.com/JuliaPlots/UnicodePlots.jl/blob/65b936860081f74d56a0946b2a83e28fd0d24f11/src/canvas.jl#L127) was not enforced before.

We now have, unless given explicit `width` and `height` or limited by `displaysize`, square `heatmap` plots and `spy` plots when inputting square matrices.

References files using square test matrices have been re-generated, I've added some checks in `tst_heatmap.jl`.


src/canvas/heatmapcanvas.jl
```diff
@@ -36,27 +36,23 @@ function printrow(io::IO, c::HeatmapCanvas, row::Int)
...
                 print(io, Crayon(foreground=fgcol), HALF_BLOCK)
             end
         else
             print(io, HALF_BLOCK)
         end
     end
-    if iscolor
-        print(io, Crayon(reset=true))
-    end
+    iscolor && print(io, Crayon(reset=true))
```
johnnychen94, line 55: I would suggest reverting style changes as you did in this file because it is not related to the PR, and would make the PR harder to review.

src/canvas.jl
```diff
@@ -108,62 +108,49 @@ function get_canvas_dimensions_for_matrix(
     width::Int, height::Int, margin::Int, padding::Int, out_stream::Union{Nothing,IO};
     extra_rows::Int = 0, extra_cols::Int = 0
 ) where {T <: Canvas}
-    min_canvheight = ceil(Int, nrow / y_pixel_per_char(T))
-    min_canvwidth  = ceil(Int, ncol / x_pixel_per_char(T))
-    aspect_ratio = min_canvwidth / min_canvheight
+    canvheight = nrow / y_pixel_per_char(T)
+    canvwidth  = ncol / x_pixel_per_char(T)
+    aspect_ratio = canvwidth / canvheight
```
johnnychen94, line 113: Maybe leave a comment here for why `aspect_ratio` is calculated this way? My understanding is that we faithfully calculate the desired aspect ratio while keeping the canvas big enough to fit all the contents?

src/interface/heatmap.jl
```diff
@@ -80,101 +80,104 @@ See also
...
-        HeatmapCanvas, nrows, ncols, max_width, max_height, width, height, margin, padding, out_stream
+
+    width, height, max_width, max_height, data_aspect = get_canvas_dimensions_for_matrix(
+        HeatmapCanvas, nr, nc, max_width, max_height, width, height, margin, padding, out_stream
     )
 
-    # ensure plot height is big enough
-    height = min(max_height, max(height, nrows))
+    # ensure plot is big enough
+    if data_aspect && aspect_ratio == 1  # square matrix should remain as is
```
johnnychen94, line 151: maybe use `aspect_ratio ≈ 1` here?

test/tst_heatmap.jl
```diff
@@ -108,3 +108,16 @@ end
...
+    seed!(RNG, 1337)
+    for m in 1:minimum(T_SZ)
+        p = @inferred heatmap(randn(RNG, m, m))
+        @test size(p.graphics.grid) == (m, m)
+    end
+    for m in minimum(T_SZ):maximum(T_SZ)
+        p = @inferred heatmap(randn(RNG, m, m))
+        s1, s2 = size(p.graphics.grid)
+        @test s1 == s2
+    end
```
johnnychen94, line 122: Just a confirm, this test breaks in master, correct?

johnnychen94 (approved):
I can understand that you have your own coding style and tastes. But committing too much style change to the history would make it difficult to track changes, it's also making others harder to review the changes. Thus I would suggest making style changes very conservatively and only doing this when you're rewriting the functionality of the related lines.

test/tst_heatmap.jl
t-bltg: Yes, for wide `displaysize` the plot widens, and the plot is not square anymore.

src/canvas.jl
t-bltg: I've added comments base on a reduced example (2x2 Matrix).

src/canvas/heatmapcanvas.jl
t-bltg: This is fixed now.

t-bltg:
As a side note, this PR on the left and master on the right:
master_pr

In `xterm`, the `heatmap(collect(1:30) * collect(1:30)')` looks more `square than this PR.

This is an artifact: querying the caret size in `xterm` with `$ printf '\e[16t` outputs `;15;7t`
`15 / 7 =  2.1428` which is different from the `HeatmapCanvas` pixel aspect ratio (`2` pixels per `y` char and `1` pixel per `x` char, which yields an aspect ratio of `2`.

t-bltg:
I'm wondering whether we should introduce a flag to disable respecting data aspect ratio ?

t-bltg:
I've added `correct_aspect_ratio` flag, which preserves the apparent squareness of a square matrix by reversing the terminal aspect ratio of `4:3`. Of course, it's preferred to keep the matrix structure in terms of characters as it's the case when `correct_aspect_ratio=false` (default).  I've added tests accordingly.

Looks like this:
spy

src/common.jl
```diff
@@ -316,7 +316,15 @@ julia_color(color)::JuliaColorType = julia_color(crayon_256_color(color))
...
 end
 
-out_stream_size(out_stream::Union{Nothing,IO}) = out_stream === nothing ? (40, 15) : displaysize(out_stream)
+# standard terminals seem to respect a 4:3 aspect ratio
+# unix.stackexchange.com/questions/148569/standard-terminal-font-aspect-ratio
+const ASPECT_RATIO = Ref(4 / 3)
+
+# default display size for the default BrailleCanvas (which has aspect ratio = 2) ==> (40, 15)
+const DEFAULT_HEIGHT = Ref(15)
+const DEFAULT_WIDTH = Ref(2round(Int, ASPECT_RATIO[] * DEFAULT_HEIGHT[]))
```
johnnychen94, line 325: Do we want to allow modifying these defaults on runtime? If not then we can use plain constants like `const ASPECT_RATIO = 4/3`.

README.md
```diff
@@ -343,6 +343,9 @@ Using a non `true monospace font` can lead to visual problems on a `BrailleCanva
 Either change the font to e.g. [JuliaMono](https://juliamono.netlify.app/) or use `border=:dotted` keyword argument in the plots.
 For a `Jupyter` notebook with the `IJulia` kernel see [here](https://juliamono.netlify.app/faq/#can_i_use_this_font_in_a_jupyter_notebook_and_how_do_i_do_it).
 
+Terminals seem to respect a standard aspect ratio of `4:3`, hence a square matrix will often no look square in the terminal.
+As a fix, you can pass `correct_aspect_ratio=true` to `spy` or `heatmap` in order to recover a unit aspect ratio.
```
johnnychen94, line 347: This seems to be an experimental flag so might be helpful to mark it as experimental.

```suggestion
(Experimental) Terminals seem to respect a standard aspect ratio of `4:3`, hence a square matrix does not often look square in the terminal.
As a fix, you can pass `correct_aspect_ratio=true` to `spy` or `heatmap` in order to recover a unit aspect ratio.
This keyword is experimental and might be unnecessary in future versions.
```

johnnychen94 (approved):
I honestly don't know where the numbers come from, but I do have tested it locally and it works well on windows terminal.

src/common.jl
t-bltg: Another reference I've found about this: https://retrocomputing.stackexchange.com/questions/5629/why-did-80x25-become-the-text-monitor-standard.

There is no modern reason for `4:3` aspect ratio, it's just historic (Fortran, punch cards, ...). This ratio was hidden before in the (40, 15) default size in `plot.jl`, I think it's better to make it explicit so that we understand how (40, 15) was computed.

README.md
t-bltg: I dislike the word `correct_aspect_ratio` it's too long. Maybe you have an idea for something shorter ? `fix_ar` ? `fix_43` ?

PR #202, opened 2022-01-16
Rework structure for `8bit` -> `24bit` mode

t-bltg:
- use `UInt32` instead of `UInt8` for canvas color (allows packing r, g, b channels for truecolor support), blending rgb, ...
- use more `Crayons` objects, reduces test files by about 25%
- toggle between `24bit` and `8bit` at runtime

- [ ] benchmark for any regressions

PR #203, opened 2022-01-17
Rework colormaps, prepare `contourplot`

t-bltg:
- Add more colormaps (8bit): rationale for this: enhanced colormap range using `8bit` colors instead of arbitrary `24bit` rgb, new reference files cover these. These new colormaps also prepare the change initiated at https://github.com/JuliaPlots/UnicodePlots.jl/pull/202 for truecolor `24bit` support.
- Change existing colormap data structure to use `Vector{NTuple}` instead of `Vector{Vector}`, enforce `const`.
- Move colorbar computations from `heatmap` to `canvas` in order to prepare `contourplot` integration, and fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/169.
- Move colormap `callback` computation to `colormaps.jl`, added tests are covering this.


src/canvas.jl
```diff
@@ -229,3 +229,48 @@ function annotate!(
     char_point!(c, char_x, char_y, text, color)
     c
 end
+
+function printcolorbarrow(
+    io::IO, c::Canvas, row::Int, colormap::Any, border::Symbol,
```
t-bltg, line 234: Arbitrary `Canvas`, for https://github.com/JuliaPlots/UnicodePlots.jl/issues/169.

src/colormaps.jl
```diff
@@ -14,1053 +14,1387 @@
...
+    :ash => _ash_data,
+    :jet => _jet_data,
 )
 
-function closestansiicolor(c)
-    tc = [0, 95, 135, 175, 215, 255]
-    _, i = findmin(abs.(c .- tc))
-    i - 1
-end
+closestansiicolor(c) = argmin(abs.(c .- UInt8[0, 95, 135, 175, 215, 255])) - 1
```
johnnychen94: Performance suggestion:

```suggestion
const _closest_ansii_color_list = UInt8[0, 95, 135, 175, 215, 255]
if VERSION >= v"1.7"
    closestansiicolor(c) = argmin(v->abs(c-v), _closest_ansii_color_list) - 1
else
    closestansiicolor(c) = argmin(abs(c-v) for v in _closest_ansii_color_list) - 1
end
```

test/references/heatmap/colormap_31x31_snow.txt
```diff
@@ -0,0 +1,15 @@
+      [90m┌───────────────────────────────┐[39m
```
johnnychen94, line 1: Where are the codes that generate these three references?

src/colormaps.jl
t-bltg: I agree on the `const`.
Couldn't we use `closestansiicolor(c) = argmin(map(v -> abs(c - v), _closest_ansii_color_list)) - 1` for both versions ?

test/references/heatmap/colormap_31x31_snow.txt
t-bltg: https://github.com/JuliaPlots/UnicodePlots.jl/blob/5beb0b6f369bb060516970bad703aab3d5959453/test/tst_heatmap.jl#L39

src/colormaps.jl
johnnychen94: `map` allocates memories; `map` almost work equivalently to `[abs(c - v) for v in ...]` and is slower than the generator version `(abs(c - v) for v in ...)`.

The v"1.7" version uses faster for-loop implementation and get rid of the generator overhead.

t-bltg: Thanks, I tested it, wasn't aware of the map(...) allocation. On `1.6`, however this fails with `ERROR: MethodError: no method matching keys(::Base.Generator`

src/plot.jl
```diff
@@ -618,7 +620,7 @@ function Base.show(io::IO, p::Plot)
         if p.show_colorbar
             print(io, plot_padding)
             printcolorbarrow(
-                io, c, row, p.colormap, p.colorbar_border, p.colorbar_lim,
+                io, c, row, callback, p.colorbar_border, p.colorbar_lim,
```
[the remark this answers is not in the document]
t-bltg: Yes, because it is enclosed in and if block: `if p.show_colorbar ...`. I've added type checks in `printcolorbarrow`.

src/colormaps.jl
t-bltg: I don't know why, but the 1.7 version `argmin(v->abs(c-v), _closest_ansii_color_list)` is completely broken and fails the CI.

t-bltg:
The CI is finally looking green.

johnnychen94 (approved):

PR #204, opened 2022-01-17
Add `contourplot`

t-bltg:
- add `contourplot` support (`levels`, `colormap`, `colorbar` controls)
- add `Contour.jl` dependency for computing levels using the [`Marching_squares`](https://en.wikipedia.org/wiki/Marching_squares) algorithm
- add `blend` flag on all canvases to plot with/without color blend (avoid overlap issues on contour lines)

https://github.com/JuliaGeometry/Contour.jl is a small requirement, since it only depends on `StaticArrays`.

See examples at https://github.com/JuliaPlots/Plots.jl/pull/4031.

The new example from `README.md` now looks like this (gaussian profile):
contourplot1

**tiny benchmark**
```julia
using UnicodePlots, BenchmarkTools

main() = begin
  x = -3:.1:3
  y = -7:.1:3
  X = repeat(x', length(y), 1)
  Y = repeat(y, 1, length(x))
  g(x, y, x₀=0, y₀=-2, σx=1, σy=2) = exp(-((x - x₀) / 2σx)^2 - ((y - y₀) / 2σy)^2)
  z = map(g, X, Y)
  @btime heatmap($z)  #   15.090 ms (197665 allocations: 9.45 MiB)
  @btime contourplot($x, $y, $z)  #   2.833 ms (29227 allocations: 1.18 MiB)
end

main()
```

Needs https://github.com/JuliaPlots/UnicodePlots.jl/pull/203.


t-bltg:
This should be easier to review now.

src/interface/contourplot.jl
```diff
@@ -0,0 +1,72 @@
...
+  `(0, 0)` stands for automatic.
+
+- **`grid`** : If `true`, draws grid-lines at the origin.
+
+Author(s)
+==========
+
+- T Bltg (https://github.com/t-bltg)
+"""
+
```
johnnychen94, line 40: extra empty line here:

```suggestion
```

src/interface/contourplot.jl
```diff
@@ -0,0 +1,72 @@
+"""
+    contourplot(x, y, z; kwargs...)
+
+Description
+============
```
johnnychen94, line 5: Does Julia's docstring system support RST style? I guess we need to follow the general documentation guideline https://docs.julialang.org/en/v1/manual/documentation/

src/interface/contourplot.jl
```diff
@@ -0,0 +1,72 @@
...
+Arguments
+=========
+
+- **`levels`** : The number of contour levels. 
+
+- **`colormap`** : The colormap to use for the contour lines. 
+
+- **`colorbar`** : Toggle the colorbar. 
+
+$DOC_PLOT_PARAMS
```
johnnychen94, line 22: Do we need to update `DOC_PLOT_PARAMS` with new item `blend`?

src/interface/contourplot.jl
```diff
@@ -0,0 +1,72 @@
...
+        collect(extrema(y)), canvas;
+        blend = blend, grid = grid, colorbar = colorbar, colormap = callback,
+        kw...
+    )
+    contourplot!(new_plot, x, y, z; name = name, levels = levels, colormap = callback)
+end
+
+function contourplot!(
+    plot::Plot{<:Canvas}, x::AbstractVector, y::AbstractVector, z::AbstractMatrix;
+    name::AbstractString = "", levels::Integer = 5, colormap = :viridis,
```
johnnychen94, line 58: Will `levels = 5` be a reasonable default here for irregular data points?

src/interface/contourplot.jl
```diff
@@ -0,0 +1,72 @@
...
+    name == "" || label!(plot, :r, string(name))
+
+    plot.colormap = callback = colormap_callback(colormap)
+    minz, maxz = extrema(z)
+
+    for cl in Contour.levels(Contour.contours(y, x, z, levels))
+        color = callback(Contour.level(cl), minz, maxz)
+        for line in Contour.lines(cl)
+            lineplot!(plot, reverse(Contour.coordinates(line))..., color=color)
+        end
```
johnnychen94, line 69: I can imagine that this can be time-consuming, so just curious, would async draw work out here?

```suggestion
        @sync for line in Contour.lines(cl)
            @async lineplot!(plot, reverse(Contour.coordinates(line))..., color=color)
        end
```

I didn't benchmark this so if it's already working quite fast, then it's great to keep it as it is.

johnnychen94 (commented):
Very nice implementation. I never used Contour before so I don't know how performant it's used in practice. Did you happen to run some real-world examples on this and ensure that it draws within a reasonable time limit?

t-bltg:
> Very nice implementation. I never used Contour before so I don't know how performant it's used in practice. Did you happen to run some real-world examples on this and ensure that it draws within a reasonable time limit?

Well in https://github.com/JuliaPlots/UnicodePlots.jl/pull/204#issue-1105915492 I've compared it against heatmap, `7x` faster and `8` times less allocs, so I'm not worried there.

I've been using it for a week withing my CFD codes (example lid driven cavity in `Plots.jl`), and it is quite fast.

src/interface/contourplot.jl
t-bltg: Dunno, I've just copied this header from `lineplot.jl` and adjusted it for `contourplot` ...

src/interface/contourplot.jl
t-bltg: Ha yes, good point.

src/interface/contourplot.jl
t-bltg: All the other plotting packages (`PyPlot`, `Plots`) use a higher number of levels, which is unreadable in `UnicodePlots`.
`5` is good since it get's you `0` level for real valued data. We can go down to `3`, but this is subjective. What be be your preference ?

src/interface/contourplot.jl
t-bltg: I'm going to test this on bigger data, but I have to think of a reference to compare to.

t-bltg: `@sync`, `@async` doesn't seem to improve the performance.

t-bltg:
@johnnychen94, quick question about drawing images: what is the convention ?
`heatmap` does not flip the `y` axis. I've added `contourplot(A)` which flips the `y` axis so that the image is not flipped up/down.

src/interface/contourplot.jl
t-bltg: I went down to 3 since our plots are really small by default.

johnnychen94:
> quick question about drawing images: what is the convention ?

From my own experience in image processing, the rotated x-y axis are used so that top-left point is the origin: https://github.com/JuliaImages/juliaimages.github.io/issues/173



Also, because the Julia array is column-major order, the first dimension (x axis) specifies the height instead of the width.

I would try to use the i-th dimension whenever the x/y notation introduces ambiguities.

src/interface/contourplot.jl
```diff
@@ -0,0 +1,59 @@
...
+    plot
+end
+
+"""
+    contourplot(A; kwargs...)
+
+Contour plot for images (flips the y axis by default)
+"""
+contourplot(A::AbstractMatrix; kwargs...) =
+    contourplot(axes(A, 2) |> collect, axes(A, 1) |> collect |> reverse, A; kwargs...)
```
johnnychen94, line 59: How about adding one extra method

```julia
contourplot(A::AbstractMatrix{CT}; kwargs...) where {CT<:Gray} = contourplot(reinterpret(eltype(CT), A); kwargs...)
```

so that `contourplot(testimage("cam"))` just works?

src/interface/contourplot.jl
```diff
@@ -0,0 +1,59 @@
+"""
+    contourplot(x, y, A; kwargs...)
+
+Draws a contour plot on a new canvas.
```
johnnychen94, line 4: I‘m really not a plot power user so it's not immediately clear to me what `x`, `y` and `A` are. Maybe rewrite the description into something like "Draws a contour plot of 2D matrix `A` along axis `x` and `y` on a new canvas."?

README.md
```diff
@@ -180,6 +181,19 @@ densityplot!(plt, randn(1000) .+ 2, randn(1000) .+ 2)
...
 ![Density Screenshot](https://github.com/JuliaPlots/UnicodePlots.jl/raw/unicodeplots-docs/doc/imgs/2.x/densityplot1.png)
 
+#### Contour Plot
+```julia
+x = -3:.1:3
+y = -7:.1:3
+X = repeat(x', length(y), 1)
+Y = repeat(y, 1, length(x))
+g(x, y, x₀=0, y₀=-2, σx=1, σy=2) = exp(-((x - x₀) / 2σx)^2 - ((y - y₀) / 2σy)^2)
+contourplot(x, y, map(g, X, Y), levels = 3)
```
johnnychen94, line 191: Does it make sense to you to also support `countourplot(x, y, g)` here? I know there's usage like `surface(x, y, g)` in Plots but I don't know how widely it's used in practice.

johnnychen94 (approved):
Clearly, you know more of the plot ecosystem than I do, so my comments here are just double checks from a user's perspective. I'll leave it to you to decide if it's ready to merge.

README.md
t-bltg: Good point, since we already support this for e.g. [lineplot](https://github.com/JuliaPlots/UnicodePlots.jl/blob/980d17cdc372599c87f88e502775d408f30fd38b/src/interface/lineplot.jl#L192). I will add this.

src/interface/contourplot.jl
t-bltg: Doesn't this require adding `Colors.jl` dependency ? Which lead to another question: should we support `colorant`s as given colors ?

johnnychen94: Oh, I see... That's a good question. If we're going to introduce ImageInTerminal encoding backend in the future, we'd eventually get those dependencies and have colorants supported...

t-bltg: What I propose is to open an issue for feature request, and we can support that later, I'll write the PR, or maybe integrate this change in https://github.com/JuliaPlots/UnicodePlots.jl/pull/202. Is is ok for you ?

t-bltg:
> From my own experience in image processing, the rotated x-y axis are used so that top-left point is the origin: JuliaImages/juliaimages.github.io#173

This means that in julia libraries, input images are stores as regular matrices (origin at North-West), which makes sense.

This is ambiguous for us since `contourplot(x, y, A; kwargs...)` uses `x` (West to East) and `y` (South to North) as in .

`PyPlot` uses this convention too: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.contour.html

t-bltg:
I think this is pretty clear to me, and from a user point of view:
```
    contourplot(A; kwargs...)

Contour plot for images (flips the y axis by default)

        Julia arrays  |  UnicodePlots
                      | 
           axes(A, 2) |      
           o------>   |   ^ 
           |          |   |
axes(A, 1) |          | y | 
           |          |   |
           v          |   o------>
                      |       x
```

johnnychen94:
Thank you again for consistently improving this package!

t-bltg:
I released it quickly so that I can move on https://github.com/JuliaPlots/Plots.jl/pull/4031.

I will rework the doc-strings now, and also add a formatting action since as as you reminded me in https://github.com/JuliaPlots/UnicodePlots.jl/pull/199,  mixing code style changes and regular changes is not a good idea for reviewing.

Issue #45, opened 2016-01-04
Improve documentation


Evizero:
I have sketched out a basic draft for the docstring documentation, but it could really use some love.

I am not very good in this sort of thing, so this kind of contribution would be very welcomed.

t-bltg:
- [ ] remove the need for screenshots in `README.md`

PR #206, opened 2022-01-19
Rework docstrings

t-bltg:
This PR automates docstring generation for arguments, function signatures, and `README.md`. Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/45 ?

Rationale:
- lots of arguments are duplicated across interfaces, with very similar signatures.
- by automating this we can check descriptions:  should start with a lowercase letter, sentences should end with a dot, ...
- removing about `200+` lines of docstring code (with duplicates) for the same result.
- simplifies maintenance for changes ( auto-generated `README.md` (default keywords), and `gen_imgs.jl` file for updating the examples).

The auto-generated `README.md` can be seen [here](https://github.com/t-bltg/UnicodePlots.jl/blob/docstring/README.md), it should be rendered close to the existing one. It didn't want to pull in `Documenter.jl`, I'm just using [`Markdown`](https://docs.julialang.org/en/v1/stdlib/Markdown/) from the stdlib to generate it dynamically.

Other changes:
- simplify sentences by aiming single phrases, and reduce string lengths (for synthetic display in `REPL`).
- follow `Markdown` syntax, as advised in `Julia`'s [documentation](https://docs.julialang.org/en/v1/manual/documentation).
- unifies `func(args..., kwargs...)` syntax.
- transform `z` -> `A` in `heatmap.jl`, since`A` reflects a matrix structure by convention.

I've carefully inspected the results of `help?> <function>` where `function` is `scatterplot`, `Plot`, `barplot`, `heatmap`, ... against master and this is either equivalent of way better structured, since the arguments now follow a fixed order. We can fine tune each function arguments by giving `desc`, `add`, or `remove` keywords for the added `arguments` or `signature` functions.

- [x] generate descriptions
- [x] generate signatures
- [x] generate `README.md` automatically
- [x] generate examples automatically

~~I'm guessing that the runtime generation of docstrings only affects precompilation of `Unicodeplots` and not load time, but I'm not sure of that.~~

~~Unexpected, and this looks bad:~~
~~**PR**~~
```
  1.353077 seconds (3.25 M allocations: 231.825 MiB, 5.12% compilation time)
```
~~**master**~~
```
  0.395560 seconds (701.22 k allocations: 41.487 MiB, 17.54% compilation time)
```
~~@johnnychen94, do you have any obvious idea why this happens ?~~

EDIT: sorry for the noise, the increased load time was just the effect of adding `Contour` dependency and thus loading `StaticArrays`.


t-bltg:
@johnnychen94, just to let you know that I think the PR is ready for review (nothing urgent though :wink: )

src/interface/barplot.jl
```diff
@@ -12,44 +11,32 @@ using a dictionary `dict`. In that case, the keys will be used as
...
-- **`xscale`** : Function or Symbol to transform the bar length before plotting.
-  This effectively scales the x-axis without influencing the captions
-  of the individual bars. e.g. use `xscale = :log10` for logscale.
-
-$DOC_PLOT_PARAMS
+# Arguments
 
-- **`symbols`** : Specifies the characters that should be used to
-  render the bars
+$(arguments(
```
johnnychen94, line 22: It's better to leave a few comments or docstrings for the internal `keywords` and `arguments`. Otherwise, future contributors would easily get confused when they see a long string interpolation like this.

docs/generate_docs.jl
```diff
@@ -0,0 +1,421 @@
+using UnicodePlots
+using Markdown
+import Markdown: MD, Paragraph, plain
+
+main() = begin
```
johnnychen94, line 5:
```suggestion
function main()
```

johnnychen94 (approved):
neat!

src/interface/barplot.jl
t-bltg: Thanks, I've added the docstring, and even better now interpolate default kwargs into description and into concrete function arguments, so that changing a default value is now dead simple.

PR #207, opened 2022-01-20
Formatter

t-bltg:
- applies minor style simplifications (`vector` ->`tuple`, `new_plot` -> `plot`, `2*x` -> `2x`, uppercase const globals, ...)
- add an automated formatting action

I don't mind which code style is applied as long as the code style is consistent.

Parameters in the added `.JuliaFormatter.toml` file follow the one we use in `Plots.jl` (see JuliaPlots/Plots.jl/pull/3749).

The formatting action is ran ~~every month (IMO weekly is too often), opening a PR if changes are detected.~~ as a CI check.


johnnychen94 (approved):
You know I'm not a big fan of committing style changes. Ideally, we should try to avoid polluting git history by including formatter commits. Instead, the formatter should be run as either CI-checks (e.g., the whitespace checks in julialang/julia), or set up locally as, for instance, the git pre-commit hooks.

But since this package is maintained by the Plots community I'll follow what you guys think is good.

t-bltg:
> Ideally, we should try to avoid polluting git history with formatter commits.

Good point. I've changed the action to run on `CI` and disabled the `CRON` job.

PR #210, opened 2022-01-22
Fix binary `or`

t-bltg:
This is a fix for https://docs.juliaplots.org/v1.25/generated/unicodeplots/#unicodeplots-ref53, where obviously a binary `or` between markers and a character from the canvas lookup table is wrong.

Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/211.


src/canvas/braillecanvas.jl
```diff
@@ -93,8 +93,9 @@ function pixel!(c::BrailleCanvas, pixel_x::Int, pixel_y::Int, color::UserColorTy
     0 <= pixel_x <= c.pixel_width || return c
     0 <= pixel_y <= c.pixel_height || return c
     char_x, char_y, char_x_off, char_y_off = pixel_to_char_point(c, pixel_x, pixel_y)
-    c.grid[char_x, char_y] =
-        Char(UInt64(c.grid[char_x, char_y]) | UInt64(braille_signs[char_x_off, char_y_off]))
+    if 0x2800 <= (val = UInt64(c.grid[char_x, char_y])) <= 0x28ff
+        c.grid[char_x, char_y] = Char(val | UInt64(braille_signs[char_x_off, char_y_off]))
+    end
```
johnnychen94, line 98: Both versions are encrypted to be... 😢

t-bltg: Sorry, I didn't get that, what versions ?

johnnychen94: The master version and the PR version... I'm afraid that I can't review this PR.

t-bltg: Oh you mean you do not understand the changes ? Let me explain: `0x2800` is [blank braille](https://unicode-table.com/en/2800/), `0x28ff` is [full braille](https://unicode-table.com/en/28FF/). Doing a `binary or` is the low level operation for 'filling' the braille pattern with dots. So I'm just checking here that the value is a braille unicode char, before applying the increment.

t-bltg: I've clarified this by using `const`s.

johnnychen94 (approved):
The result definitely looks good to me but to be honest I don't understand the details here.

PR #212, opened 2022-01-22
Ease making canvas invisible

t-bltg:
Slight followup of https://github.com/JuliaPlots/UnicodePlots.jl/pull/205, where we now use `width = 0` to make the canvas disappear (`visible` = false).

We need this to make blank plots disappear in `Plots.jl` layouts.

### DIFF
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -179,8 +179,8 @@ function Plot(
 ) where {C<:Canvas}
     length(xlim) == length(ylim) == 2 ||
         throw(ArgumentError("xlim and ylim must be tuples or vectors of length 2"))
-    length(X) == length(Y) || throw(DimensionMismatch("X and Y must be the same length"))
-    width = max(width, min_width)
+    length(X) == length(Y) || throw(DimensionMismatch("X and Y must have same length"))
+    (visible = width > 0) && (width = max(width, min_width))
     height = max(height, min_height)
 
     min_x, max_x = extend_limits(X, xlim, xscale)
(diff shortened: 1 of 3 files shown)


src/plot.jl
```diff
@@ -179,8 +179,8 @@ function Plot(
 ) where {C<:Canvas}
     length(xlim) == length(ylim) == 2 ||
         throw(ArgumentError("xlim and ylim must be tuples or vectors of length 2"))
-    length(X) == length(Y) || throw(DimensionMismatch("X and Y must be the same length"))
-    width = max(width, min_width)
+    length(X) == length(Y) || throw(DimensionMismatch("X and Y must have same length"))
+    (visible = width > 0) && (width = max(width, min_width))
```
johnnychen94, line 183: It's perhaps a matter of taste here: this version might let people wondering if `width` is still used when `visible=false`. For instance, does the following version work the same?

```julia
visible = width > 0
width = max(width, min_width)
```

t-bltg: I find it a bit counterintuitive to have `visible == false` when `width = 0` and then force a non-zero width afterwards with `max(...)`. Anyways I've checked and both versions pass our tests.

johnnychen94 (approved):

src/plot.jl
johnnychen94: > I find it a bit counterintuitive to have visible == false when width = 0 and then force a non-zero width afterwards with max(...)

Makes sense. I was thinking that we can assume `width` is not used whenever `visible==false` for entire UnicodePlots; not just this function.

t-bltg: We could also have added a `canvas_visible` kwarg to `Plot(...)`, but I though it was a bit of an overkill here.

PR #217, opened 2022-02-02
Add `surfaceplot`, `isosurface` (support 3d plots)

t-bltg:
This PR adds support for 3d plots, along with several fixes related to 3d integration.

- add `surfaceplot`
surfaceplot1
- add `isosurface`
isosurface
- support `orthographic` and `perspective` projections, `azimuth`, `elevation`, ...
- minor docstring fixes
- some small changes in `lineplot` and `scatterplot` related to the complementary `Plots.jl` PR at https://github.com/JuliaPlots/Plots.jl/pull/4089
- ~~renamed internal `show_labels` and `show_colorbar` for consistency with keywords~~
- ~~`kwargs` -> `kw`: I thought I fixed all of those in a previous commit, but inconsistencies remained, fix this for good~~
- ~~add `default_size!` to override `DEFAULT_WIDTH` and `DEFAULT_HEIGHT` in a consistent way (switch to dynamic `width` and `height` in `plot.jl`)~~

Added `NaNMath` dependency for projection related issues, [`MarchingCubes`](https://github.com/t-bltg/MarchingCubes.jl) ([registration in progress](https://github.com/JuliaRegistries/General/pull/53640)) for `isosurface`, and `StaticArrays` which was already a dependency triggered by `Contour`.


Project.toml
```diff
@@ -9,12 +9,19 @@ version = "2.7.0"
...
+MarchingCubes = "299715c1-40a9-479a-aaf9-4a633d36f717"
+NaNMath = "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
 Contour = "0.5"
 Crayons = "4.1"
+MarchingCubes = "0.1"
```
johnnychen94, line 22: Hmmm... I didn't find this package in General... What's the URL of it?

Project.toml
```diff
@@ -9,12 +9,19 @@ version = "2.7.0"
...
+NaNMath = "0.3"
+StaticArrays = "1.3"
```
johnnychen94, line 24: Is there any specific reason to restrict to StaticArrays 1.3? It seems that we only need type definitions `SVector` and `SMatrix`, so I believe older versions also works:

```suggestion
StaticArrays = "0.12, 1"
```

src/common.jl
```diff
@@ -274,8 +274,7 @@ function extend_limits(vec, limits, scale::Union{Symbol,Function})
     if mi == 0 && ma == 0
         mi, ma = map(Float64, extrema(vec))
     end
-    diff = ma - mi
-    if diff == 0
+    if abs(ma - mi) < 10eps()
```
johnnychen94, line 277:
```suggestion
    if isapprox(ma, mi; atol=10eps())
```

src/interface/isosurface.jl
```diff
@@ -0,0 +1,157 @@
...
+    if V isa Function
+        xx = repeat(x', length(y), 1)
+        yy = repeat(y, 1, length(x))
+        X = repeat(xx, 1, 1, length(z))
+        Y = repeat(yy, 1, 1, length(z))
+        Z = zero(X)
+        for (i, zi) in enumerate(z)
+            Z[:, :, i] .= zi
+        end
+        V = map(V, X, Y, Z) |> Array
```
johnnychen94, line 79: Broadcasting is simple and efficient:

```suggestion
        V = V.(y, x', reshape(z, 1, 1, length(z)))
```

```julia
function mapf(f, x, y, z)
    xx = repeat(x', length(y), 1)
    yy = repeat(y, 1, length(x))
    X = repeat(xx, 1, 1, length(z))
    Y = repeat(yy, 1, 1, length(z))
    Z = zero(X)
    for (i, zi) in enumerate(z)
        Z[:, :, i] .= zi
    end
    map(f, X, Y, Z)
end

broadcast_f(f, x, y, z) = f.(y, x', reshape(z, 1, 1, length(z)))
torus(x, y, z, r = .2, R = .5) = (√(x^2 + y^2) - R)^2 + z^2 - r^2
x, y, z = -1:.1:1, -1:.1:1, -1:.1:1

mapf(f, x, y, z) == broadcast_f(f, x, y, y) # true
@btime mapf(f, $x, $y, $z); # 40.129 μs (16 allocations: 297.25 KiB)
@btime broadcast_f(f, $x, $y, $z); # 21.685 μs (4 allocations: 72.78 KiB)
```

t-bltg:
Thanks @johnnychen94 for reviewing this !
I planned to ask for your review a bit later (when the CI would be green), but since you left some comment I will expand on them now :)

Project.toml
t-bltg: This is a package I wrote (https://github.com/t-bltg/MarchingCubes.jl). Registering is in progress (see https://github.com/JuliaRegistries/General/pull/53640), it should land this evening in `General` (mandatory waiting period of 3 days), so that we can test CI here in `UnicodePlots`. I plan to eventually transfer it to an org, maybe `JuliaGeometry` when everything will be settled here with `UnicodePlots`. By the way, comments on this code are very welcome !

Project.toml
t-bltg: Good point, I'll change that.

src/common.jl
t-bltg: Done !

src/interface/isosurface.jl
t-bltg: That's very nice (old habit from `MATLAB` and/or `Numpy` are hard to get rid of). Wait ... `x` and `y` shouldn't be permuted ! Let me have a look.

t-bltg: X and Y are permuted in 3D since we use 1st dim = x, 2nd dim = y, 3rd dim = z.
In 2D however, we use the 1st dim = y and the 2nd dim = x (plot conventions).

t-bltg:
CI is green :ok_hand: !

- [ ] improve coverage

src/interface/isosurface.jl
johnnychen94: > Maybe we should use 1st dim = z, 2nd dim = y, 3rd dim = x for 3D plots ?

I have no preference here... You can choose whatever makes sense for the broader Plots ecosystem.

src/volume.jl
```diff
@@ -0,0 +1,310 @@
...
+    lookat(center .+ cam_move, center, up_vector)
+end
+
+"""
+    MVP(x, y, z; $(keywords(; default = (), add = (:x, :y, :z, :projection, :elevation, :azimuth, :zoom, :up))))
+
+# Description
+
+Model - View - Projection transformation matrix.
+"""
```
johnnychen94, line 205: Since this is an exported symbol, it would be helpful to have a few examples and "see-also" links here so that people can quickly understand what this symbol is for. It would also be beneficial to have some links to external resources, e.g., http://www.codinglabs.net/article_world_view_projection_matrix.aspx

docs/generate_docs.jl
```diff
@@ -266,6 +288,10 @@ All plots support the set (or a subset) of the following named parameters:
 
 _Note_: If you want to print the plot into a file but have monospace issues with your font, you should probably try setting `border=:ascii` and `canvas=AsciiCanvas` (or `canvas=DotCanvas` for scatterplots).
 
+### 3D plots
+
+3d plots use a `Matrix-View-Projection` transformation matrix on input data to render 3D plots to a 2D screen. Use `elevation`, `azimuth`, `up` or `zoom` to control the `View` matrix (camera). The projection type can be set to either `:perspective` or `orthographic`. Displaying the `xyz` axes can be controlled using the `axes3d` keyword. For better resolution, use wider and taller `Plot` size.
```
johnnychen94, line 293: It's not suitable to use backquotes ` here since "Matrix-View-Projection" is not a meaningful symbol in UnicodePlots.

```suggestion
3d plots use a so-called "Matrix-View-Projection" transformation matrix `MVP` on input data to render 3D plots to a 2D screen. Use keywords`elevation`, `azimuth`, `up` or `zoom` to control the "View" matrix, a.k.a., camera. The `projection` type for `MVP` can be set to either `:perspective` or `orthographic`. Displaying the X-, Y-, and Z- axes can be controlled using the `axes3d` keyword. For better resolution, use wider and taller `Plot` size.
```

johnnychen94 (commented):
I have little knowledge on `surfaceplot` and `isosurface` but it's promising to look at the results.

For `src/volume.jl`, instead of inventing our own solution, is it possible to adopt and extend the https://github.com/JuliaGeometry/CoordinateTransformations.jl?

src/UnicodePlots.jl
```diff
@@ -59,6 +68,8 @@ export GraphicsArea,
     spy,
     boxplot,
     boxplot!,
+    draw_axes!,
```
johnnychen94, line 71: I'm not sure if we really want to export this symbol. Do users have opportunities to call this themselves instead of to call this via, e.g., `isosurface!`?

t-bltg: Yes, this was before I added the keyword `axes3d` for automating it. Will remove.

src/interface/isosurface.jl
t-bltg: The 3D convention here makes sense. I'm afraid that if we want to be consistent, I'll have to open a PR to flip axes for 2D plots. Unless it is justified by colon major access, I'll dig into that later.

src/volume.jl
t-bltg: I've added some theory links. As for the examples, there are some "plot by hand" examples (using `lineplot` and `MVP`) in `test/tst_volume.jl`.

t-bltg:
> For src/volume.jl, instead of inventing our own solution, is it possible to adopt and extend the https://github.com/JuliaGeometry/CoordinateTransformations.jl?

I've tried that road at an early stage, but couldn't get satisfactory results. I think the current is self contained and explained enough.

I still have to fix an inverse problem for the drawing of axes using a perspective transform, before we can say this PR is ready.

src/volume.jl
```diff
@@ -0,0 +1,333 @@
...
+    lookat(center .+ cam_move, center, up_vector)
+end
+
+"""
+    MVP(x, y, z; $(keywords(; default = (), add = (:x, :y, :z, :projection, :elevation, :azimuth, :zoom, :up))))
+
+# Description
+
+Build up the "Model - View - Projection" transformation matrix (see codinglabs.net/article_world_view_projection_matrix.aspx).
+"""
```
johnnychen94, line 218:
```suggestion

This is typically used to adjust how 3D plot is viewed, see also
the `projection` keyword in [`surfaceplot`](@ref), [`isosurface`](@ref).
"""
```

johnnychen94 (approved):
I don't have the bandwidth to read all the details, but this generally looks good to me. Feel free to merge when you're ready.

PR #221, opened 2022-02-07
vertical histogram base

Cvikli:
A raw VERY simple Vertical Unicode Histrogram plotting. 

I see there are a lot of work needs to be done here. If someone could be a help for me, that would be much appreciated.


t-bltg:
Thanks for this contribution.

> I see there are a lot of work needs to be done here. If someone could be a help for me, that would be much appreciated.

As a maintainer of this repository, I can push in your fork, so I will make some changes now. Don't forget to synchronize you fork accordingly using `git fetch && git pull` before you need to make changes and push new commits.

t-bltg:
I see now that you are **not** using a canvas to draw vertical histograms. This won't work, and we request you to follow the internal structure of `UnicodePlots.jl`. In `verticalhistogram` you'd need to create a `Plot` object containing a `canvas` (see https://github.com/JuliaPlots/UnicodePlots.jl#low-level-interface) on which you can draw thing. Calling `show` or `display` on that `Plot` object will print it to the screen of a file, and this is what is missing for the tests to work. I you cannot reuse one of the existing `canvas`es you'd have to create your own, but I don't think this is needed.

Cvikli:
Hey, I checked what you are saying and maybe I could make a very basic implementation. 
But sadly I won't have a single hour to finish it for 2 weeks. But after that I will try to come back!

t-bltg:
Superseded by https://github.com/JuliaPlots/UnicodePlots.jl/pull/250.

PR #223, opened 2022-02-08
Reduce allocs

t-bltg:
When doing 3D, transforming points is quite costly in terms of allocations. Maybe I'll find time to rewrite the whole transform thing by preallocating (which will certainly be a tradeoff for readability), but for now small improvements like the following are unavoidable.

**master**
```julia
julia> using BenchmarkTools, UnicodePlots
julia> @btime surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π), lines=true);
  32.321 ms (266809 allocations: 16.65 MiB)
```

**pr**
```julia
julia> using BenchmarkTools, UnicodePlots
julia> @btime surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π), lines=true);
  27.441 ms (233254 allocations: 11.25 MiB)
```

### DIFF
--- a/src/volume.jl
+++ b/src/volume.jl
@@ -283,7 +283,7 @@ function transform_matrix(t::MVP, n::Symbol)
     end
 end
 
-is_ortho(t::MVP, n::Symbol) = (
+function is_ortho(t::MVP, n::Symbol)
     if n === :user
         t.ortho
     elseif n === :orthographic
@@ -293,7 +293,7 @@ is_ortho(t::MVP, n::Symbol) = (
     else
         throw(ArgumentError("invalid n=$n"))
     end
-)
+end
 
 function (t::MVP{T})(p::AbstractMatrix, n::Symbol = :user) where {T}
     # homogeneous coordinates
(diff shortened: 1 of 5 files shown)


src/volume.jl
```diff
@@ -271,23 +271,47 @@ struct MVP{T}
...
+    if n === :user
+        t.mvp_mat
+    elseif n === :orthographic
+        t.mvp_ortho_mat
+    elseif n === :perspective
+        t.mvp_persp_mat
+    else
+        throw(ArgumentError("invalid n=$n"))
+    end
+)
```
johnnychen94, line 284:
```suggestion
function transform_matrix(t::MVP, n::Symbol)
    if n === :user
        t.mvp_mat
    elseif n === :orthographic
        t.mvp_ortho_mat
    elseif n === :perspective
        t.mvp_persp_mat
    else
        throw(ArgumentError("invalid n=$n"))
    end
end
```

src/volume.jl
```diff
@@ -271,23 +271,47 @@ struct MVP{T}
...
+    if n === :user
+        t.ortho
+    elseif n === :orthographic
+        true
+    elseif n === :perspective
+        false
+    else
+        throw(ArgumentError("invalid n=$n"))
+    end
+)
```
johnnychen94, line 296:
```suggestion
function is_ortho(t::MVP, n::Symbol)
    if n === :user
        t.ortho
    elseif n === :orthographic
        true
    elseif n === :perspective
        false
    else
        throw(ArgumentError("invalid n=$n"))
    end
end
```

johnnychen94 (approved):

PR #224, opened 2022-02-09
Interpolate color when using `lines` on `surfaceplot`

t-bltg:
Previously, the whole segment was colored using the color of the first point.
This is now way better using linear interpolation, since symmetry is preserved.
I'm not 100% happy with the implementation (a bit dirty), so this might need some rework.

- [x] reduce allocs

```julia
using UnicodePlots
surfaceplot(
    -2:2, -2:2, (x, y) -> 15sinc(√(x^2 + y^2) / π),
    elevation=80, width=80, height=30,
    zoom=1.8, colormap=:jet, lines=true
) |> display
```

**pr**
interp_pr

**master**
interp_master


src/canvas.jl
```diff
@@ -50,7 +50,9 @@ function lines!(
     y1::Number,
     x2::Number,
     y2::Number,
-    color::UserColorType,
+    c_or_v1::Union{AbstractFloat,UserColorType},  # either floating point values or colors
+    c_or_v2::Union{AbstractFloat,UserColorType} = nothing,
+    col_cb = nothing,  # callback (map values to colors)
```
johnnychen94, line 55: It's not immediately clear to me what's the purpose of these keywords. Can you explain it a bit?

johnnychen94 (commented):
The results definitely look promising and I agree that the implementation is a bit dirty.

Did you happen to try the Interpolations.jl interface? Except for an extra dependency, I assume it will be more efficient. An draft example can be found in https://juliaimages.org/ImageTransformations.jl/stable/#index_image_warping.

src/canvas.jl
t-bltg: For the old version we only needed one color (`color_or_value1` = `c_or_v1` = `a color` (int, or tuple)). Now, since I need to interpolate the color along the whole line segments, I need to pass the raw `height` values (`floats`) before mapping them to colors (since mapping to ansi 256 color is non-bijective, non-linear, so I cannot interpolate colors by using some kind of averaging). For a line I need two values `v1` and `v2` so that along the line segments, we can compute color values using `col_cb` which calls `cmapcolor(z, ...)` from `src/colormaps.jl`. Is that clearer ? If not, I will duplicate the code to make it clearer, I just wanted to reuse the code as much as possible here.

t-bltg: As for using an interpolation function from an external package I believe that this is not needed for 2 reasons: the interpolation is linear (we don't need splines, etc...) and second this is segment based so there is no `simd` computations involved, and we won't benefit from any vectorization, or whatever optimization.

t-bltg: But this PR allocates more than previously and I shall for sure fix that before merging ;)

t-bltg: `Are these shared context values for all lines!?` `lines!` is called for each line segment that you see on the image.

t-bltg: I'll have a look tomorrow, I think we need to enhance `3D` -> `2D` transformations by doing a single matmat product before going further with this.

johnnychen94: This is much clearer to me! I still have a few questions that hopefully can be answered to ensure that I get your intention correctly:

- can I assume `c_or_v1` is the color for point `(y1, x1)` and `c_or_v2` is the color for point `(y2, x2)`?
- And, for the scalar indexed-color case. How does the interpolation solve/workaround the non-linear issue?
- Why we don't do the "same" interpolation for the tuple case?

t-bltg: - yes, exactly a line going from `(x1, y1, c1)` to `(x2, y2, c2)`
- its not the interpolation, it's the order: invalid := interpolate(compute_colors(...)), valid := compute_colors(interpolate(...))
- which tuple case ? do you mean when we use colors as `RGB`s ?

t-bltg:
This is faster, less bytes allocated, but more number of allocations :/

Benchmark using the example in https://github.com/JuliaPlots/UnicodePlots.jl/pull/223.
```
julia> using BenchmarkTools, UnicodePlots
julia> @btime surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π), lines=true);
  22.277 ms (293873 allocations: 10.09 MiB)
# before (and without interpolation !) was
  27.441 ms (233254 allocations: 11.25 MiB)
```

src/canvas.jl
```diff
@@ -74,24 +138,32 @@ function lines!(
...
+        min(py, Py),
+        max(py, Py),
+        cur_x,
+        cur_y,
+        Δx / nsteps,
+        Δy / nsteps,
+    )
+
+    if c_or_v1 isa AbstractFloat && c_or_v2 isa AbstractFloat && col_cb !== nothing
+        interp_color_line!(c, args..., 1 / √(Δx^2 + Δy^2), c_or_v1, c_or_v2, col_cb)
```
johnnychen94, line 164: Is this equivalent to the following?

```julia
x0 = SVector(min(px, Px), min(py, Py))
x1 = SVector(max(px, Px), max(py, Py))
r = LinRange(x0, x1, floor(Int, nsteps))
iΔ = 1 / √(Δx^2 + Δy^2)
for i in 1:length(r)
    w = sqrt(sum(abs2, x-x0)) * iΔ
    v = (1-w) * c_or_v1 + w * c_or_v2
    pixel!(c, floor.(Int, r[i])..., col_cb(v))
end
```

t-bltg: Thanks, originally, I've split `interp_color_line!` and `const_color_line!` because of type stability issues. I'll see if this works out and simplifies.

johnnychen94: > ERROR: ArgumentError: range([0.0, 0.0], stop=[80.0, 60.0], length=1): endpoints differ

Yeah, I'm pretty sure that I didn't handle the boundary case correctly here. A shortcut is probably to check `floor(Int, nsteps) == 1`. But the final result seems incorrect...

t-bltg: Also, I'm afraid we'll allocate with things like that `cur .+= δ; all(p .≤ cur .≤ P) || continue`. Using the current loop with primitive types is a guaranty for performance.

johnnychen94: Why would you want to pass `cur_x`, `cur_y`, `δx`, `δy`, `nsteps`, `length`, `step` when you can simply pass two ranges `rx, ry`? Writing a function with 10 arguments is indeed a bad code smell to me. Compared to better names and comments, splitting this into a function doesn't make it easier to decrypt.

> I'm afraid we'll allocate with things like that cur .+= δ; all(p .≤ cur .≤ P) || continue

I'm not sure if there will be worse. The whole purpose of StaticArrays is to store arrays as tuples so that Julia compiler can optimize the allocations away.

t-bltg: `cur .+= δ` will be ok, `all(p .≤ cur .≤ P)` will allocate even if `cur` is a StaticArray, so this will be pretty bad. I'm rewriting this now with a while loop so that I can get rid of the runtime dispatch of the range arguments.

t-bltg: I've found something way simpler, and removed the ugly helper functions.

t-bltg:
It's starting to get interesting:
```julia
@btime surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π), lines=true);
  11.417 ms (73943 allocations: 4.53 MiB)
```

t-bltg:
@johnnychen94, this is when I use `SVector` and `MVector`s in loops, so this is less performant:

```julia
@btime surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π), lines=true);
  17.238 ms (141523 allocations: 6.98 MiB)
```

If you find a vectorized way of the current implementation **which keeps the allocs low**, please go ahead and push in this branch.
I probably spent too much time on this PR, so I don't plan modifying it ... Quite a lot of changes, but I think they are justified.

johnnychen94:
@t-bltg I'm currently concentrating on other projects, need a few days more to put my hands on this.

t-bltg:
```julia
@btime surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π), lines=true);
  10.975 ms (73943 allocations: 4.53 MiB)
```

t-bltg:
@johnnychen94, rebased.

johnnychen94 (approved):
The interpolation part looks okay to me. I don't have the bandwidth to read very carefully for other parts but generally nothing suspicious to me.

Given that I'm currently busy with other things, I don't want to block this PR with meaningless debates. Just want to point out that it's perhaps a good chance to revisit the function designs... a long function with >100 lines definitely gives a bad smell...

t-bltg:
Thanks for reviewing this, I totally understand being to busy for reading all the details. I'm sorry we don't have anyone else available here to comment on PRs, and since I'm a bit redundant to merge my own PRs without approval (it's kind of going against good FOSS practices), I'm systematically requesting you reviews.

> Just want to point out that it's perhaps a good chance to revisit the function designs... a long function with >100 lines definitely gives a bad smell...

I'm going to expand on this (I'm assuming you're talking about `surfaceplot!`). Long functions are sometimes necessary because of unrolling or buffering. This was the case here, since drawing interpolated lines was costly in terms of allocations and twice slower. It's much more efficient now, even if I do agree, this is a tradeoff between performance and readability. Nothing comes for free.

PR #227, opened 2022-02-11
Support vector of colors (`barplot`, `boxplot`)

t-bltg:
Fix https://discourse.julialang.org/t/unicodeplots-barplot-with-different-colored-bars/75979.


stelmo:
This is awesome, thanks 😃! Would it be possible to add the ylim (or would it be xlim) kwarg to barplot too?

t-bltg:
If the plots with distinct colors are combined in a single one, they should already be aligned, no ?

stelmo:
Yep, totally, if the plotted series are in the same graph then they should be to the same scale. The x/ylim thing would be more useful if opted to keep the plots separate, but wanted to maintain a sense of scale between them.

t-bltg:
I think using `width` is exactly what you need.

johnnychen94 (approved):

t-bltg:
> I think the keyword width is exactly what you need.

I was wrong, I've added a new keyword `upper_heigth` to reflect that, is the keyword name well chosen ?
```julia
julia> using UnicodePlots
julia> barplot(["1", "2", "3"], [1, 2, 3], upper_height = 10)
     ┌                                        ┐ 
   1 ┤■■■■ 1                                    
   2 ┤■■■■■■■ 2                                 
   3 ┤■■■■■■■■■■■ 3                             
     └                                        ┘ 
julia> barplot(["1", "2", "3", "4"], [1, 2, 3, 10], upper_height = 10)
     ┌                                        ┐ 
   1 ┤■■■■ 1                                    
   2 ┤■■■■■■■ 2                                 
   3 ┤■■■■■■■■■■■ 3                             
   4 ┤■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 10   
     └                                        ┘
```

> One last question, do these changes allow one to add a legend for bar plots, to indicate what each color means?

I've added support for `name`
```julia
p = barplot(["1", "2", "3"], [1, 2, 3], color = :blue, name = "1ˢᵗ series", upper_height = 10)
barplot!(p, ["4", "5", "6"], [6, 1, 10], color = :red, name = "2ⁿᵈ series")
```

stelmo:
Thank you very much, this looks awesome! I really appreciate all the effort you put into this request :) Once the package I am developing to visualize some flux measurements at the repl for [COBREXA.jl](https://github.com/LCSB-BioCore/COBREXA.jl) is done, I will ping you so that you can see what came of it :) Once again, thank you very much!!!

t-bltg:
I will release a new version when https://github.com/JuliaPlots/UnicodePlots.jl/pull/224 is merged.

Issue #232, opened 2022-02-20
Option to hide tick labels


juliohm:
Sometimes it is useful to hide the tick labels with `xticks=false` or `yticks=false` or even set the labels manually. I am opening the issue in case the feature is welcome.

t-bltg:
To hide the labels I think you are looking for `labels = false`.
As for the setting the value manually, you can use either `xlim` or `ylim`, if I understood correctly.

juliohm:
Thank you for the great additions!

t-bltg:
> Ok, I'll have a look in the next few days or so.

Woops, I lied :laughing:.

PR #234, opened 2022-02-21
Support `xticks` - `yticks`

t-bltg:
Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/232.


src/description.jl
```diff
@@ -62,6 +64,8 @@ const DESCRIPTION = (
     xlim = "plotting range for the `x` axis (`(0, 0)` stands for automatic)",
     ylim = "plotting range for the `y` axis",
     zlim = "colormap scaled data range",
+    xticks = "draw xticks",
+    yticks = "draw yticks",
```
johnnychen94, line 68:
```suggestion
    xticks = "set `true` to draw ticks on x-axis, default value is `true`.",
    xticks = "set `true` to draw ticks on y-axix, default value is `true`.",
```

johnnychen94 (approved):
Pretty straightforward to me.

Issue #255, opened 2022-05-15
Plots as string


FedeClaudi:
Hey, 

Awesome job with Unicode Plots, love the package. 

I'd like to have a simple way to get a string representation of the plots to use them with `Term.jl` (see https://github.com/FedeClaudi/Term.jl/issues/86). 

```Julia
string(histogram(randn(1000)))
```

almost works, but the resulting string doesn't include (ANSI ?) color codes so that printing it gives a normal looking plot but with no color. The simplest way I've found is to use an `IOBuffer` to capture a call to print:
```Julia
"""
Turn a UnicodePlots.Plot into a String
"""
function capture_plot(plot)
    io = IOBuffer()
    print(IOContext(io, :color => true), plot)
    return String(take!(io))
end
```

but that's a bit clunky, it would be great to have an easier way to turn a plot into a string and keep the color info. 

Thanks!

PR #257, opened 2022-05-15
stringify plot

t-bltg:
Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/255.

```julia
using UnicodePlots
p = lineplot(1:2)
s = string(p; color=true)
```


t-bltg:
@FedeClaudi, does that fix your issue ?

FedeClaudi:
Looks like it would! Awesome. That was so quick :)

PR #259, opened 2022-05-17
code quality

t-bltg:
After reading https://discourse.julialang.org/t/discussion-on-why-i-no-longer-recommend-julia-by-yuri-vishnevsky/81151/67 :wink:

### DIFF
--- a/src/canvas.jl
+++ b/src/canvas.jl
@@ -34,7 +34,7 @@ function points!(
 ) where {T<:UserColorType}
     length(X) == length(Y) == length(color) ||
         throw(DimensionMismatch("X, Y, and color must be the same length"))
-    for i in 1:length(X)
+    for i in eachindex(X)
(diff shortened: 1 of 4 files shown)


src/canvas.jl
```diff
@@ -34,7 +34,7 @@ function points!(
 ) where {T<:UserColorType}
     length(X) == length(Y) == length(color) ||
         throw(DimensionMismatch("X, Y, and color must be the same length"))
-    for i in 1:length(X)
+    for i in eachindex(X)
```
johnnychen94, line 37: Because here we don't have `@inbounds` so it's quite "safe"

johnnychen94 (approved):

src/canvas.jl
t-bltg: I know, juste being picky on good practices :)

PR #260, opened 2022-05-22
lazily load FreeTypeAbstraction

johnnychen94:
The story here:

- for users that don't export png files, they get 0.8s latency reduced
- for users that need, nothing is changed. -- the FreeTypeAbstraction loading is delayed to the first `png_image` call

Note: **LazyModules could possibly introduce world-age issues if not used appropriately**. Luckily here FreeTypeAbstraction is a perfect black box -- both `renderstring!` and `findfont` outputs something we already know, i.e., String, Matrix.

Before:

```julia
  1.676243 seconds (6.64 M allocations: 485.373 MiB, 4.00% gc time, 11.02% compilation time: 72% of which was recompilation)

julia> p = @time lineplot([0, 1]);
  0.429121 seconds (601.23 k allocations: 40.566 MiB, 2.89% gc time, 99.79% compilation time)

julia> p = @time lineplot([0, 1]);
  0.000054 seconds (99 allocations: 11.016 KiB)

julia> @time savefig(p, "foo.png");
  2.858724 seconds (6.03 M allocations: 417.951 MiB, 4.60% gc time, 74.08% compilation time: 18% of which was recompilation)

julia> @time savefig(p, "foo.png");
  0.269588 seconds (37.30 k allocations: 7.499 MiB)
```

After:

```julia
  0.880895 seconds (4.09 M allocations: 306.096 MiB, 3.19% compilation time: 5% of which was recompilation)

julia> p = @time lineplot([0, 1]);
  0.406931 seconds (585.60 k allocations: 39.381 MiB, 99.81% compilation time)

julia> p = @time lineplot([0, 1]);
  0.049308 seconds (38.73 k allocations: 2.581 MiB, 99.79% compilation time: 48% of which was recompilation)

julia> @time savefig(p, "foo.png");
  4.058094 seconds (9.49 M allocations: 651.377 MiB, 4.35% gc time, 68.28% compilation time: 5% of which was recompilation)

julia> @time savefig(p, "foo.png");
  0.642027 seconds (660.92 k allocations: 50.671 MiB, 5.24% gc time, 54.63% compilation time: 99% of which was recompilation)
```

---

This doesn't improve TTFX performance, only "reduces" the package loading latency.
It's perhaps a good idea to use this trick in large umbrella packages such as Plots and Makie so let me ping a few people here @jkrumbiegel @SimonDanisch @BeastyBlacksmith @lorenzoh @ChrisRackauckas


t-bltg:
On second calls:

> julia> p = @time lineplot([0, 1]);
  0.049308 seconds (38.73 k allocations: 2.581 MiB, 99.79% compilation time: 48% of which was recompilation)

or

> julia> @time savefig(p, "foo.png");
  0.642027 seconds (660.92 k allocations: 50.671 MiB, 5.24% gc time, 54.63% compilation time: 99% of which was recompilation)

Why does it recompile ?

johnnychen94:
Oh, interesting. I actually don't know much about the internals so I could be wrong. But let me try to explain it: is that world-age changes (every time we load a bunch of new methods, e.g., loading a package, the world age increases) trigger invalidation and thus recompilation.

```julia
  0.901052 seconds (4.09 M allocations: 306.083 MiB, 3.52% compilation time: 5% of which was recompilation)

julia> p = @time lineplot([0, 1]);
  0.418570 seconds (585.60 k allocations: 39.381 MiB, 99.80% compilation time)

julia> p = @time lineplot([0, 1]);
  0.000063 seconds (99 allocations: 11.016 KiB)

julia> @time savefig(p, "foo.png");
  4.212479 seconds (9.49 M allocations: 651.379 MiB, 4.44% gc time, 68.22% compilation time: 5% of which was recompilation)

julia> @time savefig(p, "foo.png");
  0.657460 seconds (660.92 k allocations: 50.671 MiB, 5.23% gc time, 55.35% compilation time: 99% of which was recompilation)

julia> @time savefig(p, "foo.png");
  0.280837 seconds (37.63 k allocations: 7.526 MiB)
```

The first call triggers the package (FreeTypeAbstraction) loading, invalidates a few methods, then the second call recompiles it. Things get normal when it comes to the third call -- because no more packages are loaded in the second call.

t-bltg:
> Things get normal when it comes to the third call

Good, I was afraid the timings were showing a regression.

t-bltg (approved):

johnnychen94:
The overhead is about 60-80ns in my machine, so should be nothing to worry about for common non-trivial operations.

```julia
julia> using Colors, BenchmarkTools, LazyModules

julia> @lazy import ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
LazyModule(ImageCore)

julia> @btime rand(RGB);
  3.219 ns (0 allocations: 0 bytes)

julia> @btime rand(ImageCore.RGB);
  75.468 ns (1 allocation: 32 bytes)
```

johnnychen94:
This PR becomes a perfect example to cut off the package loading latency using the LazyModules trick so let's wait for a few days to raise some discussion (or, Q&A). I'd like to see how other people think about this.

johnnychen94:
The even interesting story here is that UnicodePlots itself is also a black box, so it's doable for downstream packages to lazily load UP as well, for instance, https://github.com/JuliaImages/DitherPunk.jl/pull/77 switches from Requires to LazyModules

src/UnicodePlots.jl
```diff
@@ -13,7 +15,7 @@ import MarchingCubes
 import NaNMath
 import Contour
 
-import FreeTypeAbstraction
+@lazy import FreeTypeAbstraction = "663a7486-cb36-511b-a19d-713bb74d65c9"
 import ColorTypes
 import FileIO
```
t-bltg, line 20:
```suggestion
@lazy import ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
@lazy import FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
```

johnnychen94: Handling world age issues needs a bit of practice, but in principle, you'll very much likely hit it when you try to directly process the data if its type comes from the lazily loaded package.
Thus for `ColorTypes.RGB` and `MarchingCubes.MC`, they are unknown alien types unless we load them. -- This is also why the world age issues disappear when you try to call the same function again because the first failure actually triggers package loading, so no longer alien types.

---

I won't worry too much about MarchingCubes because it contributes very little to the loading latency using the julia-1.9 dev utility `@time_imports`.

```julia
julia> @time_imports using UnicodePlots
      0.6 ms  ┌ Compat
      0.9 ms  ┌ LazyModules
      4.6 ms    ┌ OrderedCollections
     42.4 ms  ┌ DataStructures
      0.4 ms  ┌ NaNMath
     22.4 ms  ┌ FixedPointNumbers
    302.2 ms  ┌ StaticArrays
     44.5 ms    ┌ ChainRulesCore
     45.7 ms  ┌ ChangesOfVariables
      0.7 ms    ┌ InverseFunctions
      3.4 ms    ┌ DocStringExtensions 67.89% compilation time
      3.3 ms    ┌ IrrationalConstants
      0.7 ms    ┌ DataAPI
      0.3 ms    ┌ StatsAPI
      0.3 ms    ┌ SortingAlgorithms
      0.7 ms    ┌ LogExpFunctions
      5.5 ms    ┌ Missings
     36.1 ms  ┌ StatsBase 14.35% compilation time
      1.0 ms  ┌ ConstructionBase
      0.7 ms  ┌ MarchingCubes
     30.7 ms  ┌ Crayons
    152.9 ms  ┌ Unitful
      1.3 ms  ┌ Contour
    107.9 ms  ┌ ColorTypes 4.23% compilation time
    809.8 ms  UnicodePlots 3.90% compilation time
```

t-bltg:
@johnnychen94, it looks like recent addition of `LazyModules` fails `Plots.jl` with the `UnicodePlots` backend (`FileIO` related, export as png). Those tests are only ran on `ubuntu` which explain why `MacOS` and `Windows` are not failing.

See https://github.com/JuliaPlots/Plots.jl/runs/6595180678?check_suite_focus=true.

Can you take a look ?

johnnychen94:
IIUC this is due to the same reason why LazyModules doesn't apply to MarchingCubes -- DataFormat is an alien type to the world-age until FileIO is loaded.

My best guess is that FileIO needs to provide a better black box interface, e.g., `save(:png, io, img)` using Symbol. Otherwise, we should probably just roll back to the eager version.

t-bltg:
Thanks for fixing.

PR #269, opened 2022-06-18
fix plot size in `png` - rework docs

t-bltg:
- fix incorrect plot size when exporting to png (buggy `polarplot`);
- cache font for speeding up `png_image` & `savefig`;
- add banner for pending `Term` 1.0.


t-bltg:
We need to cache the font since we hit https://github.com/JuliaGraphics/FreeTypeAbstraction.jl/issues/67.

src/canvas/dotcanvas.jl
```diff
@@ -9,17 +9,13 @@ DOT_DECODE[4] = ':'
...
-Similar to the `AsciiCanvas`, the `DotCanvas` only uses
-ASCII characters to draw it's content. Naturally,
-it doesn't look quite as nice as the Unicode-based
-ones. However, in some situations it might yield
-better results. Printing plots to a file is one
-of those situations.
+Similar to the `AsciiCanvas`, the `DotCanvas` only uses ASCII characters to draw it's content.
+Naturally, it doesn't look quite as nice as the Unicode-based ones.
+However, in some situations it might yield better results.
+Printing plots to a file is one of those situations.
```
waldyrious, line 15: Just dropping a note to commend you for using semantic line breaks instead of a fixed width :)

t-bltg: Thanks, I didn't know that it was called that way, it just felt more readable ;)

waldyrious: It's actually a [best practice](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) that goes back all the way to the early Unix days :)

t-bltg: 1974 ! Wasn't even born, dully noted :+1:

Issue #264, opened 2022-05-28
Auto-resize based on terminal size


zsz00:
Layout with Term example get a chaotic display, auto resize would be a great feature to have:

Windows Terminal     v 1.12.10982.0
```julia
julia> versioninfo()
Julia Version 1.8.0-beta3
Commit 3e092a2521 (2022-03-29 15:42 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: 96 × Intel(R) Xeon(R) Platinum 8255C CPU @ 2.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, cascadelake)
  Threads: 1 on 96 virtual cores
```

```julia
julia> using UnicodePlots, Term

julia> panel(plot; kw...) = Panel(string(plot, color=true); fit=true, kw...)
panel (generic function with 1 method)

julia> print(
         panel(lineplot([cos, sin], -π/2, 2π); title="lineplot", style="yellow") *
         panel(contourplot(-3:.01:3, -7:.01:3, (x, y) -> exp(-(x / 2)^2 - ((y + 2) / 4)^2)); title="contourplot", style="red") *
         panel(surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π)); title="surfaceplot", style="blue") / (
           panel(histogram(randn(1_000_000), nbins=100, vertical=true); title="histogram", style="green") *
           panel(densityplot(randn(1_000), randn(1_000)); title="densityplot", style="cyan")
         )
       )

```

![image](https://user-images.githubusercontent.com/12711073/170813028-fb23238d-d936-4eb2-ac4a-cc435ee2ea08.png)
![image](https://user-images.githubusercontent.com/12711073/170813042-59bd26be-d56b-492c-aed3-5473a5d8facc.png)

t-bltg:
Thanks for reporting. I believe it is a `Term` bug and we should therefore ping @FedeClaudi ...

FedeClaudi:
Hi, 

What's happening there is that the whole layout is larger than the width of your terminal so each single line continues in the next, breaking the visuals ( you can see on the left of the second image the end of the blue panel on the right. Resizing your terminal window should fix this. 

[...]

PR #271, opened 2022-06-19
`densityplot`: support `dscale` - cleanup

t-bltg:
Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/226.

@exaexa can you test if this suits you ?

PR #272, opened 2022-06-20
UnicodePlots `3.0.0`

t-bltg:
I feels more natural in julia to store the canvas with `rows` -> `plot height` and `cols` -> `plot width`. However, since we print plots row by row in a terminal, it's inefficient to iterate over the columns of a `grid` row.

The previous solution was to use `grid[width, height]`, but you have to mentally transpose the canvas for writing algorithms. I solve this by transposing the grid on creation yielding `grid[height, width]` so that we can index using `grid[row, col]` and retain the performance when iterating over columns of a given row (printing). `nrows(c)` mapping to `size(c.grid, 1)` is more easy mentally.

Thus, high level functions `lines!`, `points!`, `pixel!`, `char_point!`, `annotate!` still retain the `(x, y)` order in their arguments, but the grid and colors access is now swapped: `grid[y, x]` being more intuitive.

Follow up of https://github.com/JuliaPlots/UnicodePlots.jl/pull/267.
Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/264.

- [x] Add `Changelog` file
- [x] Add `Developer Notes` section

All changes:
- swap `(w, h)` on canvas `grid` and `colors` - breaking change for canvas creation;
- rename `printrow` to `print_row` (breaking since exported);
- remove deprecated declarations and corresponding tests (breaking);
- unexport canvas related functions (low level, should use `import UnicodePlots: lines!, ...`);
- rename `zscale` to `dscale` for `densityplot` (ok since https://github.com/JuliaPlots/UnicodePlots.jl/pull/271 not yet released);
- huge (ridiculous) performance improvement for displaying colormap canvases on stdout by buffering `i/o` (see ex below);
- add `array` keyword to `heatmap` for matrix display in array convention (origin at NW, `y` pointing downwards, `x` pointing to the right);
- remove functor support (if haven't seen any real example of usage, and is only partially supported);
- allow `lineplot(x::Vector, y::Matrix)` for plotting multiple series mapping to matrix columns;
- allow `width = :auto` and `height = :auto` for creating a plot based on the current terminal size;
- add `ColorSchemes.jl` dependency, and remove all hard-coded colormap tables (no performance impact);
- reduce number of exported symbols to high level methods;
- add `head_tail_frac` for `lineplot` using `head_tail`;
- add `hline!` and `vline!`;
- remove `spy` default title (redundant);
- more code simplification / unification.

```julia
main() = display(heatmap(collect(1:150) * collect(1:150)', width = 150, height = 150))

main()  # compile
@time main()
1.812467 seconds (994.36 k allocations: 35.468 MiB)  # before (no buffering)
0.134729 seconds (635.64 k allocations: 25.404 MiB)  # pr (~14 faster)
```


t-bltg:
@johnnychen94, I'm pinging you since these are breaking changes and will require a major bump.
I think there are very few people working with explicit `Canvas`es instead of `Plot`s and using `printrow`, if any, meaning that these changes should be transparent to the majority of the end users.
The deprecation warnings have been there for a long time so it's time to remove them from the code.

src/canvas/densitycanvas.jl
```diff
@@ -68,25 +69,25 @@ function pixel!(c::DensityCanvas, pixel_x::Int, pixel_y::Int, color::UserColorTy
...
 
-function printrow(io::IO, print_nc, print_col, c::DensityCanvas, row::Int)
+function print_row(io::IO, _, print_color, c::DensityCanvas, row::Int)
     0 < row ≤ nrows(c) || throw(ArgumentError("Argument row out of bounds: $row"))
     signs = DEN_SIGNS[]
     fact = (length(signs) - 1) / c.max_density[]
-    for x in 1:ncols(c)
-        val = fact * c.zscale(c.grid[x, row])
-        print_col(io, c.colors[x, row], signs[round(Int, val, RoundNearestTiesUp) + 1])
+    for col in 1:ncols(c)
```
johnnychen94, line 93: maybe `@inbounds`?

src/common.jl
```diff
@@ -178,7 +178,7 @@ const FSCALES = (identity = identity, ln = log, log2 = log2, log10 = log10)  # f
 const ISCALES = (identity = identity, ln = exp, log2 = exp2, log10 = exp10)  # inverse
 const BASES = (identity = nothing, ln = "ℯ", log2 = "2", log10 = "10")
 
-const CRAYONS_FAST = Ref(true)
+const CRAYONS_FAST = Ref(false)
```
johnnychen94, line 202: What's the reasoning behind this change?

johnnychen94 (commented):
Sorry for the delay! I only checked the 1) swap `x`-`y` and 2) IO buffer part. It looks good! For other changes, I assume you know more details than me, and I know you have the commitment to fix whatever breaks afterward.

I notice there are a lot of breaking changes here, maybe it's worth adding a CHANGELOG.md file and keeping a few noteworthy changes there so that users are aware of it when they find things breaking.

Regards to performance tweak, IndirectArrays.jl is perhaps a good concept to utilize. If we can reform the idea "print color by color" to "getindex an array element by element", a lot of things can be made easier and cleaner -- the buffer IO can even be replaced by a buffer vector (which can be more efficient to use and reuse). But this requires a heavy rewrite on `print`, `print_row`, `print_color` stuff and I'm not 100% certain that this applies to UnicodePlots (I'm trying to refactor XTermColors in this way).

Re `x`-`y` swap, we might need to draw a clear line on user interface and implementation details. A figure like this in README with a few words on it could be helpful to understand the codes. For example,

"""
Developer note on the axis convention. Because Julia uses column-major indexing order for an array type, UnicodePlots internally treats the vertical axis as the "x"-axis for performance consideration. Thus there will be a manual conversion from the user interface convention "x-y" order to the internal implementation "y-x" order (by `transpose`-ing the array). Common users of UnicodePlots don't need to be aware of this axis difference if sticking to public interface.
"""

image

src/common.jl
t-bltg: Woops, my bad (development), will remove.

src/canvas/densitycanvas.jl
t-bltg: good point, will add.

t-bltg:
Thanks for taking the time to comment !

> I notice there are a lot of breaking changes here, maybe it's worth adding a CHANGELOG.md file

:+1: I'll add a `CHANGELOG.md`.

> Regards to performance tweak, IndirectArrays.jl is perhaps a good concept to utilize. If we can reform the idea "print color by color" to "getindex an array element by element", a lot of things can be made easier and cleaner -- the buffer IO can even be replaced by a buffer vector (which can be more efficient to use and reuse). But this requires a heavy rewrite on print, print_row, print_color stuff and I'm not 100% certain that this applies to UnicodePlots (I'm trying to refactor XTermColors in this way).

Isn't any kind of indirect access preventing optimizations ?

> Re x-y swap, we might need to draw a clear line on user interface and implementation details. A figure like this in README with a few words on it could be helpful to understand the codes.

Good idea, I'll add that, and we can even draw the figure with `UnicodePlots` (with `x-y` swapped, since you are describing the old buffer ;)).

t-bltg:
Regarding the colors, do you think we should set `USE_LUT` to `true` by default (i.e. don't use 256 colors for named colors, but true colors) ? I kept this to `false` before to retain compatibility, but I'm not sure this will be better for end users ...

t-bltg:
I will probably merge this PR tomorrow.

Issue #292, opened 2022-09-20
`DateTime` ticks not disappearing when setting `xticks=false` option


ghyatzo:
Hello,

when I tried plotting a timeseries using DateTime objects as the time, passing the `xticks=false` keyword does nothing and the bottom left and bottom right labels persists. 

The only way I found is using the `label!` function passing in an empty string but it is a bit hacky.

t-bltg:
Could you post a [mwe](https://en.wikipedia.org/wiki/Minimal_reproducible_example) ?

ghyatzo:
```
using DataFrames
using Dates
using UnicodePlots

df= DataFrame(Time=DateTime[], Test=Float64[])
for i = 1:100
	t = now()
	push!(df, [t, rand(1:100)])
	sleep(0.1)
end
lineplot(df.Time, df.Test; xticks=false, yticks=false)
```
gives me this
![Screenshot 2022-09-20 at 11 46 32](https://user-images.githubusercontent.com/8601724/191226038-8508aa62-3b39-4c96-a57c-e46bc6b90ec2.jpg)

PR #293, opened 2022-09-20
`Dates`: fix `xticks` - add `format`

t-bltg:
Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/292.

@ghyatzo, I've also added the ability to control the ticks format using the new argument `format` (see https://docs.julialang.org/en/v1/stdlib/Dates/#Dates.format-Tuple{TimeType,%20AbstractString}).

Example usage: https://github.com/JuliaPlots/UnicodePlots.jl/blob/32cee13c7274add4fa80d583d5c328951782c827/test/tst_lineplot.jl#L75

I myself do not use `DataFrames`, so please comment if you need anything else.


ghyatzo:
No this is perfect. Thank you!

PR #310, opened 2022-11-13
README: Fix two section headings

waldyrious:
- Typo: "Know issues" → "Known issues"
- Normalize section heading capitalization to Sentence case

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,7 +466,7 @@ julia> Pkg.add("UnicodePlots")
   ```
 </details>
 
-##### Know Issues
+##### Known issues
 <details>
   <summary></a><b>...</b></summary><br>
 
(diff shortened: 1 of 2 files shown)


t-bltg (approved):

t-bltg:
@waldyrious I'm starting to think of a `Documenter` based docs, would that make more sense ?

waldyrious:
Possibly... but I confess I've been little more than a lurker in Julia for the past several years now (and in particular I haven't experimented playing with Documenter.jl as I wished I had by now), so I don't really think my opinion holds much weight here :)

t-bltg:
Thanks, I'll open up an issue about that.
Using only a single `README.md` has some limitations, and we start to see those.

PR #333, opened 2023-02-10
Clarify `xticks`, `yticks` in README.md

Socob:
Just a tiny `README.md` change, but since there aren’t any actual ticks (i. e. little marker lines on the axes), I think it makes sense to clarify that `xticks`/`yticks` actually hide the tick **labels**.


t-bltg:
Good point, thanks.

Can you modify and run `$ cd docs; julia gen_docs.jl` instead ?
`README.md` is auto-generated.

Socob:
Oh whoops, didn’t realize that. Will do!

t-bltg:
Oh no wait, it's `src/description.jl` that should be modified.

😅

Socob:
@t-bltg No worries, that’s what I understood it to mean. All done now!

t-bltg:
Awesome, thanks!

Issue #335, opened 2023-02-22
[FR] try to remove the `StaticArrays` dependency


t-bltg:
Benchmark the impact of `SMatrix` usage over regular matrices, and decide whether or not we can get rid of the `StaticArrays` dependency.

PR #337, opened 2023-02-23
remove `StaticArrays` dependency

t-bltg:
Fix https://github.com/JuliaPlots/UnicodePlots.jl/issues/335.

At first glance this doesn't seem to impact performance, but we should find other use cases to compare:
```julia
julia> @time begin
    using UnicodePlots
    display(surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π), lines=true))
end

# master
3.770246 seconds (5.82 M allocations: 383.574 MiB, 5.44% gc time, 13.51% compilation time: 6% of which was recompilation)

# PR
3.553505 seconds (5.89 M allocations: 385.593 MiB, 5.43% gc time, 12.98% compilation time: 5% of which was recompilation)
```

I find a bit hard though ~~to introduce type instability in helpers such as `autolims`~~, and increase allocations by using regular vectors.

EDIT: using `Val` trying to fix the added invalidations, it doesn't seem to help :/

Hum, this PR seems to impact the reference images, probably because of `ldiv` different algorithms. But local tests are passing :shrug: .


Moelf:
is your timing on 1.8.5 or 1.9 beta? one thing I commented earlier is they way we do precompilation here is not very efficient, and thus the loading time is too long, I believe StaticArrays is part of the problem but because I've seen StaticArrays.jl alone takes 1+ s, I'm surprised the time saved here is only 0.2 s

t-bltg:
The timings are on `1.9.0-beta4`, yeah there are some fluctuations I should probably use `BenchmarkTools` instead, breaking down (too bad `@time_imports` is broken on `1.9.0-beta4`):

**3.4.0**
```julia
  3.297305 seconds (5.49 M allocations: 361.066 MiB, 5.83% gc time, 3.89% compilation time: 24% of which was recompilation)
julia> @time display(surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π), lines=true))
  0.416089 seconds (322.86 k allocations: 21.905 MiB, 98.01% compilation time)
```

**pr**
```julia
  3.249613 seconds (5.57 M allocations: 364.484 MiB, 6.78% gc time, 3.66% compilation time: 18% of which was recompilation)
julia> @time display(surfaceplot(-8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π), lines=true))
  0.422238 seconds (314.46 k allocations: 20.044 MiB, 98.50% compilation time)
```

And the reason is ... that https://github.com/JuliaGeometry/MarchingCubes.jl loads `StaticArrays` :laughing:, I totally forgot that: this is endless.

Moelf:
ah yes... all makes sense now, I was about to say that due to excessive (?) compilation I'm worried that even if we remove StaticArrays, we may end up shifting loading time into loading pkgimg...

t-bltg:
The problem with compilation time is that how can we decide which plotting function is more important than another and should go into the precompile scenario ?

Also maybe serialization and de-serialization of `pkgimages` code would enhance in future julia versions ?

Moelf:
>which plotting function is more important than another and should go into the precompile scenario

the ideal case is we don't have to precompile every single possible plots, but we can use snoopcompile to identify the most common 20% of low-ish level functions and precompile those

PR #341, opened 2023-04-03
Use relative paths when including extensions in __init__

andreasnoack:
Currently, the paths are resolved when the module is compiled which breaks relocatability.


t-bltg:
Thanks !

PR #375, opened 2024-03-20
try loading package last in tests

IanButterworth:
May be the cause of https://github.com/JuliaLang/julia/issues/53688#issuecomment-2009888495


t-bltg:
Superseded by https://github.com/JuliaPlots/UnicodePlots.jl/pull/376.

PR #437, opened 2026-04-29
Use original values for x/y labels for custom scaling functions

pjaap:
Hey there, thanks for this amazing package!

We started to use it a lot in our projects and I discovered a minor bug, which should be fixed by this PR.

If a custom scaling function is used, then the range labels are wrong: they are also passed through the scaling function.

MWE: 
image

You see, the y-labels should NOT be 200 and 10_000.

with the fix here, it looks correct:

image

Same result for the `xscale`.


t-bltg:
Thanks for the catch, I think you are right, and the scaling function shouldn't modify labels.

Il have slightly reworked the code and added regression tests.

pjaap:
Thanks for polishing and merging!

PR #440, opened 2026-06-08
Don't use BrailleCanvas for vertical histogram by default

topolarity:
The vertical histogram uses Unicode Block Elements (U+2580–U+259F) for most of its diagram, but it inherits its "blank space" Braille (U+2800) character from the active canvas.

In the terminal this works fine since everything is monospace, but in browsers this often causes visual mis-alignment since, e.g., on macOS the default fonts give Braille characters a different spacing.

Before:
```
       ┌          ┐
   130  █⠀⠀⠀⠀⠀⠀⠀⠀▇
        █⠀⠀⠀⠀⠀⠀⠀⠀█
        █⠀⠀⠀⠀⠀⠀⠀⠀█
        █⠀⠀⠀⠀⠀⠀⠀⠀█
        █⠀⠀⠀⠀⠀⠀⠀⠀█
        █⠀⠀⠀⠀⠀⠀⠀⠀█
        █⠀⠀⠀⠀⠀⠀⠀⠀█
        █⠀⠀⠀⠀⠀⠀⠀⠀█
        █▄⠀⠀⠀⠀⠀⠀▄█
        ██▂⠀⠀⠀⠀▃██
        ███▇▆▅▇███
        ██████████
        ██████████
        ██████████
     0  ██████████
       └          ┘
       ⠀0⠀⠀⠀⠀⠀⠀⠀⠀2⠀
       μ ± σ: 1.0 ± 0.71
```

After:
```
       ┌          ┐
   130  █        ▇
        █        █
        █        █
        █        █
        █        █
        █        █
        █        █
        █        █
        █▄      ▄█
        ██▂    ▃██
        ███▇▆▅▇███
        ██████████
        ██████████
        ██████████
     0  ██████████
       └          ┘
        0        2
       μ ± σ: 1.0 ± 0.71
```

The remaining mis-alignment issues in the browser are due to Github, Slack, etc. using a `font-family` that causes some glyphs to be filled in by one font and others to be provided by another. We can't do much about that, but at least we can fix this


topolarity:
`pre` failures are because the `hash` implementation in Julia 1.13, and `Contours.jl` seeds its algorithm with the `first(...)` entry of a dict (the contour lines changed, so the plot did as well)

Not related to this PR anyway

t-bltg:
Oh I see you are right, vertical histogram have slightly diverged from the original implementation of horizontal histograms which are using `BlockCanvas`es.

I have to cleanup the implementation at some point but your fix is very much valid.

t-bltg (approved):

PR #450, opened 2026-09-18
Add GeoInterface extension

evetion:
So you can plot most geometries across the Julia ecosystem.

Screenshot 2026-09-17 at 22 46 44

```jl

using UnicodePlots, GeoInterface, NaturalEarth
const GI = GeoInterface

# NaturalEarth feature collections
countries = naturalearth("admin_0_countries", 110)
rivers = naturalearth("rivers_lake_centerlines", 110)
places = naturalearth("populated_places", 110)

kw = (; width=80, height=20, grid=false)
p = geoplot(countries; name="countries", title="World", kw...)
geoplot!(p, rivers; name="rivers", color=:blue)
geoplot!(p, places; name="places", color=:red)
display(p)
```

I used Claude Opus.


t-bltg:
This looks very neat and promising !

Can you add some basic tests ?

I don't know why CI is failing (seems unrelated), if you could look at (and hopefully fix it) it'd be awesome.

evetion:
Thanks! I will add tests. Regarding the current CI, it seems that some reference checks are failing, not everything is 100% identical. Although not all platforms/versions.

evetion:
> I don't know why CI is failing (seems unrelated), if you could look at (and hopefully fix it) it'd be awesome.

Claude gives the following, seems we should fix it in a separate PR. Main is failing too.

1. tst_contourplot.jl — "padding" and "consistency with surfaceplot" (Julia 1 and pre, all OSes, 24-bit step). Contour lines are off by a pixel here and there. Julia-version dependent: same machine, same Contour 0.6.3 and ColorSchemes 3.31.0, passes on 1.10, fails on 1.13. That matches CI exactly (lts green, 1/pre red).
2. tst_issues.jl — "isolated colorbar (#169)" (8-bit step, ubuntu only; that's why lts-ubuntu is red while lts-macos/windows are green). Caused by Crayons 4.1.1 → 4.2.0, which rewrote to_256_colors to pick the nearer of the grey ramp and the 6×6×6 cube instead of the old heuristic. Pinning Crayons 4.1.1 locally reproduces the stored reference byte-for-byte; with 4.2.0 six mid-range colorbar rows differ.

So updating Crayons in CI should be enough for 2, but for 1 we should choose between Julia 1.10 and 1.13 (or have seperate references for both?).

t-bltg:
Ok, thanks for investigating with AI, we should regenerate ref imgs for Julia 1.13+ I guess, and for Crayons, I have to take a look at the offending PR.

EDIT : the offending PR is https://github.com/KristofferC/Crayons.jl/pull/70.

t-bltg:
I have fixed CI for `Crayons` 4.2 in https://github.com/JuliaPlots/UnicodePlots.jl/pull/451, so we can ignore the `lts` failures now.

With CI fixed, we only need a few tests now ...

UnicodePlots/test/runtests.jl
```diff
@@ -136,34 +137,37 @@ macro no_allocs(ex)
 end
 
 println("\n== start: testing with $(UnicodePlots.colormode())bit colormode ==\n")
-Crayons.force_color(true)  # withenv("FORCE_COLOR" => "X") do  # JuliaPlots/UnicodePlots.jl/issues/134
```
t-bltg, line 139: Can you leave these lines unchanged ?

t-bltg (approved):

